PR #12, opened 2016-11-26
fixed bug causing inputs to only recieve single space deliminated tokens

### LINKED ISSUE #9: Answers can only be one word
Sorry for the flood of issues. Having a lot of fun with this lib though. :smile:

Using the demo in the README, it seems that the survey.Input prompt truncates answers to just the first word. (Try using something like "hello dolly" as the answer.)

I'm on a roll with something, or I'd submit a patch right now -- maybe sometime in the future if you can't get around to it, however I'm not super familiar with terminal-fu.

AlecAivazis:
Fixes #9 by changing the input prompt to accommodate for intended behavior: https://github.com/golang/go/issues/5703

### DIFF
--- a/input.go
+++ b/input.go
@@ -1,8 +1,11 @@
 package survey
 
 import (
+	"bufio"
+	"errors"
 	"fmt"
 	tm "github.com/buger/goterm"
+	"os"
 
 	"github.com/alecaivazis/survey/format"
 )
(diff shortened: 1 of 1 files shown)


AlecAivazis:
@mholt would you mind verifying that this works for someone else?

mholt:
@AlecAivazis Works for me!

PR #14, opened 2016-11-26
Add ability to ask just one question (closes #8)

### LINKED ISSUE #8: A simple way to ask one question at a time
I have a situation where a fair amount of logic needs to happen between each question, and it would be good to be able to do something like this:

```go
answer, err = survey.AskOne(&survey.Choice{
	Message: "...",
	Choices: []string{"..."},
})
```

What do you think?

mholt:
Sorry didn't have a chance to write tests. But take a look and see what you think.

### DIFF
--- a/survey.go
+++ b/survey.go
@@ -23,6 +23,22 @@ type Prompt interface {
 	Cleanup(string) error
 }
 
+// AskOne asks a single question without performing validation on the answer.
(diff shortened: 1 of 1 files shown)


AlecAivazis:
LGTM - merging. Thanks again for taking the time to help me find those edge cases / add some missing features!

PR #16, opened 2016-11-27
AskOne now returns an error

AlecAivazis:
@mholt - just a heads up on a slight deviation from what you submitted in your PR. If you have a strong opinion as to why the empty string is better, i'm all ears!


mholt:
I forewent the error for convenience. If there is an error in getting an answer with AskOne, I don't really care since AskOne doesn't validate the answer anyway. But I don't really have a preference. :)

PR #25, opened 2017-04-01
use templates for output formatting control

coryb:
Hi Alec,

The goal of this PR is to allow users to easily control the output for the prompts by using go's `text/template` library.  The templates allow for color injection as well as basic formatting.

The PR is larger than I had hoped, so sorry for that.  

Thanks for `survey` it is just what I was looking for!  I am hoping to add Confirm and Checkbox type prompts similar to Inquire as well.

If you have any questions or concerns please let me know, or if you would prefer to go in a different direction that is fine too :)

Thanks!
-Cory


AlecAivazis:
Hey Cory, thanks for the awesome PR - I have been trying to find the time for this exact refactor and it has been slipping through the cracks. I'm realllly excited to see the `Confirm` and `Checkbox` prompts too.

In general, this looks great - really makes the various prompts more maintainable. I'm also a big fan of the `➤` character for choices, however in my terminal it does not appear to lineup with the question mark. I know this is pretty minor, but it's driving my OCD slightly crazy. Do you mind aligning those? It looks like there is an extra space before the selected marker. Here's a picture for reference

screen shot 2017-04-01 at 5 08 09 pm

coryb:
Ok, yeah, my bad.  For some reason I thought the original implementation had 3 spaces there with the arrow in the middle, so I missed it, will fix now.

coryb:
I am working on Checkbox now, will submit that in separate PR.

AlecAivazis:
Awesome! I'm going to merge this now - thanks again!!

PR #30, opened 2017-04-03
add Confirm prompt

coryb:
Adding basic Y/n confirmation prompt.  The implementation is similar to MultiChoice in that I have added an Answer that users can populate to get a `bool` result, otherwise it returns a `Yes` or `No` string depending on the answer.  This will also loop if we get an answer that we cannot parse.

```
$ go run ./examples/confirm.go
? Are you happy? (Y/n) Yeah
✘ Sorry, your reply was invalid: "Yeah" is not a valid answer, try again
? Are you happy? Yes
response string: Yes
response happy: true
```

### DIFF
--- a/confirm.go
+++ b/confirm.go
@@ -1,0 +1,131 @@
+package survey
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"regexp"
+
+	tm "github.com/buger/goterm"
+)
+
+// Confirm is a regular text input that accept yes/no answers.
+type Confirm struct {
+	Message string
+	Default bool
+	Answer  *bool
+}
+
+// data available to the templates when processing
+type ConfirmTemplateData struct {
+	Confirm
+	Answer string
+}
+
+// Templates with Color formatting. See Documentation: https://github.com/mgutz/ansi#style-format
+var ConfirmQuestionTemplate = `
+{{- color "green+hb"}}? {{color "reset"}}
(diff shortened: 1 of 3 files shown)


AlecAivazis:
Hey Cory, thanks for yet another PR. I'll leave some comments inline.

I did some thinking about our problem today and what do you think about extending the analogy with `json.Unmarhsal` and changing the API of `Ask` to receive a pointer to a struct where the values for the entire survey is stored. We could use field tags to link up a question name with the field if we didn't want to enforce some conventions. This allows us to avoid the whole type problem to begin with from the consumer's perspective. It doesn't really affect this PR (since it makes 100% sense for `Confirm` to return a boolean), but I wanted to get your thoughts.

confirm.go
```diff
@@ -0,0 +1,131 @@
...
+		if err != nil {
+			return "", err
+		}
+		// send the message to the user
+		fmt.Print(out)
+		return confirm.Prompt()
+	}
+
+	// return the value
+	*confirm.Answer = answer
```
AlecAivazis: What if the user doesn't provide an answer when instantiating this prompt?

coryb: doh! I meant to add that logic, forgot.  Like in MultiChoice I just check to see if the value is `nil` and then create a default answer.

AlecAivazis:
Awesome! Thanks for making those changes. I'm going to merge this in now.

Issue #31, opened 2017-04-04
Modify API to allow for non-string result from prompts?


coryb:
@AlecAivazis, creating an issue to track discussions started in [#28] and [#30] regarding potential API changes to encapsulate variable data returned from prompts (example: `MultiChoice` could return an `[]string` and `Confirm` could return a `bool`).

Thinking about your last suggestion, if I understand what you are saying the usage would look roughly like:
```
    var simpleQs = []*survey.Question{{
        Name: "name",
    }, {
        Name: "color",
    }, {
        Name: "days",
    }, {
        Name: "happy",
    }}

    answers := struct {
        Name  string
        Color string
        Days  []string
        Happy bool
    }{}

    err := survey.Ask(simpleQs, &answers)
```

Or more with tags:
```
    answers := struct {
        UserName   string   `survey:"name"`
        FavColor   string   `survey:"color"`
        PreferDays []string `survey:"days"`
        IsHappy    bool     `survey:"happy"`
    }{}

    err := survey.Ask(simpleQs, &answers)
```

How would we extend `AskOne` to work on interface values as well?  Perhaps something like:
```
days := []string{}
err := survey.AskOne(prompt, &days)
```

I think the usage of this is good from a user perspective.  However I think the implementation will be reasonably tricky though (ie lots of code) and we will need to do lots of reflection (which will potentially move some compile-time errors to runtime-errors).  So I think it is do-able but non-trivial, and I am not sure the implementation complexity cost would justify the relative ease of use in the API.

coryb:
If we continue with the `Answer` field pattern in the prompt structs, I think we could achieve similar usage above, something like:
```
    answers := struct {
        UserName   string
        FavColor   string
        PreferDays []string
        AmHappy    bool
    }{}

    var simpleQs = []*survey.Question{{
        Prompt: &survey.Input{
            Answer: &answers.UserName,
        },
    }, {
        Prompt: &survey.Choice{
            Answer: &answers.FavColor,
        },
    }, {
        Prompt: &survey.MultiChoice{
            Answer: &answer.PreferDays,
        },
    }, {
        Prompt: &survey.Confirm{
            Answer: &answers.AmHappy,
        },
    }}

    err := survey.Ask(simpleQs)
    fmt.Printf("Answers: %v", answers)
```

The `AskOne` usage could look something like:
```
    p := &survey.Confirm{}
    err := survey.AskOne(p)
    fmt.Printf("Answer: %v", p.Answer)
```
or
```
[...]

PR #32, opened 2017-04-15
Added support for windows

AlecAivazis:
This PR adds support for both `cmd.exe` and `PowerShell`.

@coryb if you have the time to take a look and make sure I didn't cause any regressions in your recent changes, that would be great! There is still some work to do before I merge this in but I wanted to get this out early since I'm pretty sure I finished the prompts you added.


core/terminal.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// SoundBell uses the default Bell() behavior in order to not lose functionality.
+func (t *Terminal) SoundBell() {
+	fmt.Fprintf(t, "%c", readline.CharBell)
+	t.Bell()
+}
+
+// DevNull is a writer that writes no where (ie, leave the printing up to me)
+type DevNull struct{}
```
coryb: probably should be `ioutil.Discard`? https://golang.org/pkg/io/ioutil/#pkg-variables

AlecAivazis: Definitely - nice catch

coryb:
Hey, I dont fully grok all the changes, but it looks like things are headed in a good direction.  One minor thing that concerns me is making the templates (and template data types) private.  That will prevent users (ie me) from changing the prompts when desired (perhaps I want to change the colors or styling).

The use of `github.com/k0kubun/go-ansi` looks promising, at least a lot of the cursor movement logic looks much cleaner.   I was starting to play with `github.com/gosuri/uilive` which I thought looked promising for eliminating all the cursor logic as well (basically treats the prompt as a print buffer, every time you flush you refresh the prompt).

AlecAivazis:
Ah good point about the templates, I'll make them public. 

Also `go-ansi` has served pretty well in my testing but there are some bugs that I have fixed locally in it's windows support and still need to get them incorporated into the original project. `github.com/gosuri/uilive` looks pretty interesting. Let me know how that goes! 

Once this gets merged in, i'm going to start working on the api we discussed in #31.

AlecAivazis:
@coryb it looks like your name is missing from the Contributors tab on GitHub, not sure if you care but that probably means you're missing some configuration on your local machine to tie your commits to your GH account

coryb:
Thanks for the heads up about the contributors.  Not a big deal, but I was committing with my work git config, it should be remedied the next time I send you a PR :)

PR #35, opened 2017-04-18
fix incompat issues between posix and windows EraseInLine arguments

coryb:
the posix EraseInLine expected values 0-2, while the windows EraseInLine expected values 1-3.  To resolve the issue I just named several constants to match the posix expected values then in the switch statement for windows use those constants.

From the Wikipedia entry on CSI-n-K usage:
> Erases part of the line. If n is zero (or missing), clear from cursor to the end of the line. If n is one, clear from cursor to beginning of the line. If  n is two, clear entire line. Cursor position does not change.

I have also changed all of the EraseInLine usage to clear the entire line, not just from cursor.  The functionality currently will be the same but if the cursor ever moves during prompting (like some work I am planning on doing) then erasing the entire line is necessary, and the "correct" thing to do in any case I think.

### DIFF
--- a/terminal/display.go
+++ b/terminal/display.go
@@ -1,11 +1,9 @@
-// +build !windows
-
 package terminal
 
-import (
-	"fmt"
-)
+type EraseLineMode int
 
-func EraseInLine(mode int) {
-	fmt.Printf("\x1b[%dK", mode)
-}
+const (
+	ERASE_LINE_END   EraseLineMode = iota
+	ERASE_LINE_START EraseLineMode = iota
+	ERASE_LINE_ALL   EraseLineMode = iota
+)
--- a/terminal/display.go
+++ b/terminal/display.go
@@ -3,7 +3,7 @@ package terminal
 type EraseLineMode int
 
 const (
-	ERASE_LINE_END   EraseLineMode = iota
-	ERASE_LINE_START EraseLineMode = iota
-	ERASE_LINE_ALL   EraseLineMode = iota
+	ERASE_LINE_END EraseLineMode = iota
+	ERASE_LINE_START
+	ERASE_LINE_ALL
 )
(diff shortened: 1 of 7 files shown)


terminal/display_posix.go
```diff
@@ -0,0 +1,11 @@
+// +build !windows
+
+package terminal
+
+import (
+	"fmt"
+)
+
+func EraseInLine(mode EraseLineMode) {
```
AlecAivazis: So this name is an artifact from `go-ansi` and I think it would probably be more appropriately named something like `EraseLine` - thoughts?

terminal/display.go
```diff
@@ -1,11 +1,9 @@
...
-	"fmt"
-)
+type EraseLineMode int
 
-func EraseInLine(mode int) {
-	fmt.Printf("\x1b[%dK", mode)
-}
+const (
+	ERASE_LINE_END   EraseLineMode = iota
+	ERASE_LINE_START EraseLineMode = iota
```
AlecAivazis: Sorry to be a stickler, but do you mind removing the repeated `EraseLineMode = iota?

AlecAivazis:
Fantastic, i just verified that a known bug is gone on Windows!

terminal/display_posix.go
coryb: Sorry confused, you want the func `EraseInLine` renamed to `EraseLine` or the `EraseLineMode` renamed to `EraseLine`?   I think the function name could go either way.  The CSI-n-K operator is apparently called "EL – Erase in Line" so I think that is where the name came from.

AlecAivazis: Yea sorry, could have been more clear. I do see the benefit of sticking close to the old API but I think i prefer `EraseLine`. The "In" feels very repetitive unnecessarily

coryb: okay, updated func name

terminal/display.go
coryb: np, removed redundant noise.

AlecAivazis:
Perfect, thanks again!

PR #40, opened 2017-04-24
use a common render function to write all the prompts

coryb:
Hey @AlecAivazis, the goal of this PR is to isolate the prompt rendering and subsequent refresh and cursor movement into a common routine.  I did this by embedding a common `renderer` object into each prompt.  I couldn't find another way that made any sense. Overall I think it is a simplification although `Password` became more complicated due to the way `readline` works.  `readline` really wants to manage the prompt and tries to do a bunch of tricky cursor manipulation with no obvious ways to work around it.  I spent a lot of time fighting with `readline`.

This PR also merges the `Question` and `Choice` templates into a single template so that we can have common render logic for `Select` and `MultiSelect`.

I migrated the tests to be more table driven, which simplifies them, I also started using `github.com/stretchr/testify/assert` which I have found to be very useful.

Lastly I fixed a minor bug with the HideCursor, where if you ctrl-c the prompt the ShowCursor was not getting called.  So I wrapped that in a defer to always call ShowCursor even if interrupted.

Anyway, check it out and let me know if you have any questions or concerns.  Thanks!

### DIFF
--- a/core/renderer.go
+++ b/core/renderer.go
@@ -1,10 +1,9 @@
-package survey
+package core
 
 import (
 	"io/ioutil"
 	"strings"
 
-	"github.com/AlecAivazis/survey/core"
 	"github.com/AlecAivazis/survey/terminal"
 	"github.com/chzyer/readline"
 )
@@ -12,7 +11,7 @@ import (
 // basic readline config that does not put terminal in Raw mode
 // so that basic editing works (backspace, arrow keys etc).  This
 // is used by all text-input prompt types
-var simpleReadlineConfig = &readline.Config{
+var SimpleReadlineConfig = &readline.Config{
 	Stdout: ioutil.Discard,
 	FuncMakeRaw: func() error {
 		return nil
@@ -24,12 +23,12 @@ var simpleReadlineConfig = &readline.Config{
 	EOFPrompt:    "\n",
 }
 
-type renderer struct {
+type Renderer struct {
 	lineCount int
 	mask      rune
 }
 
-func (r *renderer) render(tmpl string, data interface{}) error {
+func (r *Renderer) Render(tmpl string, data interface{}) error {
 	// clean out current line in case tmpl didnt end in newline
 	terminal.EraseLine(terminal.ERASE_LINE_ALL)
 	// clean up what we left behind last time
@@ -39,7 +38,7 @@ func (r *renderer) render(tmpl string, data interface{}) error {
 	}
 
 	// render the template summarizing the current state
-	out, err := core.RunTemplate(tmpl, data)
+	out, err := RunTemplate(tmpl, data)
 	if err != nil {
 		return err
 	}
--- a/password.go
+++ b/password.go
@@ -1,17 +1,14 @@
 package survey
 
 import (
-	"bytes"
-	"fmt"
-
-	"github.com/AlecAivazis/survey/terminal"
 	"github.com/chzyer/readline"
+
+	"github.com/AlecAivazis/survey/core"
 )
 
 // Password is like a normal Input but the text shows up as *'s and
 // there is no default.
 type Password struct {
-	renderer
 	Message string
 }
 
@@ -20,49 +17,9 @@ var PasswordQuestionTemplate = `
 {{- color "green+hb"}}? {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }} {{color "reset"}}`
 
-type filteredReadlinePasswordOutput struct {
-	len int
-}
-
-// Write will be called by readline to echo the masked characters to the console.  It assumes
-// full control of the output with prompting, but since survey is printing the prompt readline
-// does not play nice here.  Write is called with clear-line statements which need to be
-// ignored then it is called with the full buffered output. So if you type "hi" this is the
-// sequence that will be sent to Write
-// 0: "\033[J\033[2K\r"
-// 1: " \b"
-// 2: "\033[J\033[2K\r"
-// 3: "*" (this is the masked "h" from "hi")
-// 4: "\033[J\033[2K\r"
-// 5: "**" (this is the masked "hi" from "hi")
-func (o *filteredReadlinePasswordOutput) Write(p []byte) (n int, err error) {
-	// we need to ignore the reset sequence unless we have already printed masked
-	// characters in which case we need to move the cursor back for however many
-	// characters we have printed previously
-	if bytes.Equal(p, []byte("\033[J\033[2K\r")) {
-		if o.len > 0 {
-			// we have previously printed, so move back some number of characters
-			// then clear to end of line
-			terminal.CursorBack(o.len)
-			terminal.EraseLine(terminal.ERASE_LINE_END)
-		}
-		return 0, nil
-	}
-
-	// when the buffer is empty readline.RuneBuffer.output will print
-	// this sequence to move the cusror forward then backwards
-	// we dont want to count these so just print and return
-	if bytes.Equal(p, []byte(" \b")) {
-		return fmt.Fprintf(readline.Stdout, "%s", p)
-	}
-
-	o.len = len(p)
-	return fmt.Fprintf(readline.Stdout, "%s", p)
-}
-
 func (p *Password) Prompt(rl *readline.Instance) (line interface{}, err error) {
 	// render the question template
-	err = p.render(
+	out, err := core.RunTemplate(
 		PasswordQuestionTemplate,
 		*p,
 	)
@@ -72,9 +29,10 @@ func (p *Password) Prompt(rl *readline.Instance) (line interface{}, err error) {
 
 	// a configuration for the password prompt
 	config := rl.GenPasswordConfig()
-	config.Stdout = &filteredReadlinePasswordOutput{}
+	// use the right prompt (make sure there is an empty space after the prompt)
+	config.Prompt = out + " "
+
 	config.MaskRune = '*'
-	rl.SetConfig(config)
 
 	// ask for the user's Password
 	pass, err := rl.ReadPasswordWithConfig(config)
--- a/password.go
+++ b/password.go
@@ -1,14 +1,17 @@
 package survey
 
 import (
-	"github.com/chzyer/readline"
+	"bytes"
+	"fmt"
 
-	"github.com/AlecAivazis/survey/core"
+	"github.com/AlecAivazis/survey/terminal"
+	"github.com/chzyer/readline"
 )
 
 // Password is like a normal Input but the text shows up as *'s and
 // there is no default.
 type Password struct {
+	renderer
 	Message string
 }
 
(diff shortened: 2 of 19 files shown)


AlecAivazis:
Hey @coryb thanks for yet another PR. I will try to make my way through this tomorrow, i'm having a bit of trouble focusing tracking the changes related to a single one of these features. I'm not sure how tightly coupled they are but if it would be possible to split this into a few different PRs that would be a huge help. 

I want to make sure we can get your stuff in as soon as possible and don't want a large PR to stall the rest of the improvements. My first reaction is that the move to test based tables is a huge improvement, i like the unification of the rendering-refresh logic but I haven't gotten a chance to grok the password changes. 

Thanks again!

coryb:
Hey, yeah, sorry it is another big PR, I just worked my way through all the various issues I ran into and it was bigger than I hoped. Most of it is a bit coupled, so I am not sure how I can pull them apart.  I can submit one for the HideCursor bit, that is stand-alone.  I can also submit one for just one prompt type (perhaps `Input` with the renderer logic), plus the testing refactor for that prompt.  I can leave the other prompt types out for now to keep it simple.

AlecAivazis:
well i suppose I see three separate parts: the `HideCursor` fix, the refactor of the tests to tables, and the renderer refactor. I'm totally fine with having all of the inputs as one, just want to try to think about this renderer addition without worrying about holding back the other improvements

coryb:
okay, I will see what I can do.

coryb:
Okay, I have created #41 and #42 for the `HideCursor` issue and table driven tests.  Assuming they are merged to master I will then merge them into this PR.

AlecAivazis:
I just saw, thanks so much for doing that! Make it much easier for me to merge them with confidence. I will review the two tonight assuming I dont fall asleep soon and tackle the third tomorrow

coryb:
Okay, cool, I am going to pass out myself, so I will update this PR in the morning if the others get merged.  Thanks!

terminal/print.go
```diff
@@ -5,21 +5,21 @@ import (
 )
 
 var (
-	ansiStdout = NewAnsiStdout()
+	AnsiStdout = NewAnsiStdout()
```
coryb: fyi, made this public so I could override it to capture output for testing.

coryb:
Okay, updated the PR with the merge from master.  That should should help reduce some of the noise.

coryb:
btw, I noticed you removed most of the examples, just wondering why.  I found the extra examples useful for testing changes like these, so I wrote them all again, but I have not checked them in.

AlecAivazis:
@coryb sorry it's taken me so long to go through this PR. I'm in the final stretch before a big release at work so all of my focus is there. That being said, I'm starting to go through it now. 

I moved your examples to the test directory which is where I have been putting all of the integration tests and edge cases that have been reported. Is there one missing?

coryb:
Hey, no problem, work is useful for paying the bills.  

Doh, I missed the examples in the tests directory, that explains it :)

AlecAivazis:
Haha no worries, I haven't really found a really satisfying way to prevent regressions as we work on this package. Keeping a directory of all of the different issues and examples people have provided was the simplest thing for now. 

 I meant for the example dir to be more consumable snippets of code for common use cases. Although I'm not sure it's really that useful now that I've fleshed out the README

coryb:
I suppose we would use `expect` to automate prompt and response testing in a quasi automated fashion.  If I get some free time I will play with it.

AlecAivazis:
Wow, i hadn't heard of `expect` before, this looks really promising! and https://github.com/ThomasRooney/gexpect looks like a great way to keep external dependencies to a minimum

password.go
```diff
@@ -17,9 +20,49 @@ var PasswordQuestionTemplate = `
...
+// does not play nice here.  Write is called with clear-line statements which need to be
+// ignored then it is called with the full buffered output. So if you type "hi" this is the
+// sequence that will be sent to Write
+// 0: "\033[J\033[2K\r"
+// 1: " \b"
+// 2: "\033[J\033[2K\r"
+// 3: "*" (this is the masked "h" from "hi")
+// 4: "\033[J\033[2K\r"
+// 5: "**" (this is the masked "hi" from "hi")
+func (o *filteredReadlinePasswordOutput) Write(p []byte) (n int, err error) {
```
AlecAivazis: Hmm so I still have to go through this a bit more but I wonder if it's better to just leave password as sort of an odd man and remove the `renderer` from its implementation. There is a massive win in unifying all of the re-rendering logic but I fear this might cause more bugs than the pain in having to potentially change two different places for rendering logic.

terminal/print.go
AlecAivazis: Yea i noticed, nice idea - since the `Ansi` bit is a bit of an artifact of the old package, how do you think about just naming this `Stdout`? I know I probably make some dumb naming request on every PR (sorry!) but the API is already a bit verbose, i think it would be good to try to minimize it wherever possible

password.go
coryb: I understand your concern, we could do that.   The main reason for this refactor for me is to be able to generate `help` messages in the prompt directly.  So my current plan is to have something like a `Help string` property for each Prompt which will allow me to provide hints to prompts like:
``` 
    prompt := &survey.Input{
        Message: "Enter birthday:",
        Help: "Birthday format must be in Day/Month/Year",
    }
```
Then the prompt would look something like this by default:
```
? Enter birthday: [? for help] 
```
then if user enters `?` we rewrite the prompt as:
```
Birthday format must be in Day/Month/Year
? Enter birthday: 
```
So if we make Password the odd-one-out it would also not be able to used to participate in the Help messages.  I would more confident about leaving it out if we had a strategy/plan to pull it back in but short of purging readline, not sure what else we could do.

coryb:
I think `gexpect` is cheating, you need to get familiar with the up-and-coming language `tcl` with expect :)

terminal/print.go
coryb: okay, np, the naming is pretty arbitrary to me.  Really we should probably replace all the `fmt.Print*(...)` calls with `fmt.Fprint*(Stdout, ...)` so we can test/capture all the output, but that is probably a change for later.

coryb: updated

password.go
AlecAivazis: There you go again coming up with ideas that make me question the interface. 

I really like this and it feels like something that a question should be concerned with instead of a prompt but i don't really like the idea of passing `helpText` to `Prompt` although it would closer resemble the signature of `Cleanup`. 

Is this something you imagine on multi-selects too? I'm not sure how one would collect a request for help in the various multi-line prompts. In the case of non-`Input` prompts, i'm not sure how I feel about showing help text based on an input the user cannot see they typed. FWIW, Inquirer somewhat addresses in multi-line prompts by showing a short help text describing the arrow key behaviour until you press one.

AlecAivazis: Also, i'm not opposed to removing readline - i did it to remove the old logic I had in this package to listen to `/dev/tty` which always made me uneasy even though windows seemed to support it. So long as we can keep cross platform support, anything is on the table if it improves the package.

AlecAivazis: Second also, maybe it's worth removing the `renderer` from `Password` and we can try to pull that logic into somewhere we can use in both places, i agree with you that if this would totally preclude a password prompt from the help text feature, we need to think hard about the next step

coryb: If help text is not directly associated with the prompt I am not sure how it could be integrated.  It is necessarily contextual to the question asked by the prompt.  FWIW, I have been doing this with some fugly in-house code (that I am trying to replace with survey) and have had very positive user feedback.

Yeah, I would do it for select and multi-select also, we would just capture the '?' key-press like we capture the arrow keys and space key (which are also inputs the user cannot see).  I have played with the implementation a bit and I think it will be okay (and intuitive).  Without setting `Help` it will be identical to the current behavior.

coryb: okay, I can revert Password.  I like that `readline` works on windows, which is why I have not tried to replace it, I didnt want to sign up to guarantee windows support on any other implementation :)  Perhaps there are simpler alternatives for readline that more closely fit what we want, or maybe I just make a readline-lite that just does the bits we like.

AlecAivazis: Yea it's probably better to just leave it up to the prompt to "decide" if it wants to support help text. 

I'm definitely interested in the feature, just want to make sure we think through all the possibilities so we don't end up having to bump a major version because we missed something obvious. I suppose it fits in the same way `Message` is defined on the prompt. Although, I might be able to see an argument to move that to the question type too, but let's not go too far down that rabbit hole just yet.

As a side note, I think it would be good to show some help text to describe the arrow key behaviour of the multiline inputs for exactly this reason - was planning on adding it soon

coryb:
Password changes reverted.

renderer.go
```diff
@@ -0,0 +1,55 @@
+package survey
```
AlecAivazis: Okay, last thing before we get this in - do you mind moving this inside of `core`? I know its not very conventional, but I tend to prefer internal packages with public exports to having a large flat codebase with internal types/variables.

AlecAivazis:
I saw, I just finished going through the code, going to take the tests for a spin and then we should be able to get this in before we go to sleep!

renderer.go
coryb: moved it to core, had to make a bunch of stuff public since it is no longer in the same package.

AlecAivazis: thanks - running the last of the tests now

coryb:
I am going to crash now, if you want more changes I will jump on it in the morning.  Thanks for spending the time on this!

AlecAivazis:
All the tests pass except for `ask.go` and `input.go` which both need to be updated to match the new API

password.go
AlecAivazis: Well ever since adding windows support to survey and getting a hang of the native windows libraries, it seemed sort of reasonable to just build a cross platform `renderer` which you sort of started already. might be worth a shot adding the necesary things to `terminal` to completely replace `readline` with `core.renderer`

core/renderer.go
```diff
@@ -0,0 +1,54 @@
...
+	"strings"
+
+	"github.com/AlecAivazis/survey/terminal"
+	"github.com/chzyer/readline"
+)
+
+// basic readline config that does not put terminal in Raw mode
+// so that basic editing works (backspace, arrow keys etc).  This
+// is used by all text-input prompt types
+var SimpleReadlineConfig = &readline.Config{
```
AlecAivazis: Is there a reason to use this config instead of the one provided by default in `terminal.GetReadline()` which should return a configured instance that is already passed though - i figured you needed a different config here than in the default one but it's not clear to me why

coryb: Yeah, the config from `terminal.GetReadline()` still puts the tty into raw mode which prevents  the default terminal echo.  The `SimpleReadlineConfig` leaves the tty in raw mode by stubbing out the Enter/ExitRawMode callbacks.  Without the `SimpleReadlineConfig` readline does all the output control which erases my prompt via `033[J` in the runebuffer (this is the stuff I tried to work around in Password) .  The only other way to keep the readline-rawmode and get this to work is to implement an OnChange function that basically re-implements the echo functionality from your tty that is disabled by readline.  I went that way first but found that I would also have to re-implement all the native tty functionality for editing (ie backspace, arrow keys to type over content didnt work).

AlecAivazis: ahhh gotcha. thanks for the explanation - i knew you would have a good reason

coryb:
Fixed ask.go and input.go tests to pass in new required param which is required when intializing the Prompt structs without named fields.

PR #42, opened 2017-04-24
refactor output tests to be more table driven

coryb:
This also includes 2 whitespace changes to Select and MultiSelect to fix dangling space at end of prompt.

### DIFF
--- a/confirm_test.go
+++ b/confirm_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/AlecAivazis/survey/core"
+	"github.com/stretchr/testify/assert"
 )
 
 func init() {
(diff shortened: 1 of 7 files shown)


confirm_test.go
```diff
@@ -4,75 +4,44 @@ import (
...
-	)
-	if err != nil {
-		t.Errorf("Failed to run template to format input answer: %s", err)
-	}
-
-	expected := `? Is pizza your favorite food? Yes`
-
-	if actual != expected {
-		t.Errorf("Formatted input answer was not formatted correctly. Found:\n%s\nExpected:\n%s", actual, expected)
+func TestConfirmRender(t *testing.T) {
```
AlecAivazis: Do you mind adding a `Name` to the table rows so we can provide some context when these tests fail?

coryb: Okay, added a `title` field for each test row.  On failure you will now see something like:
```
--- FAIL: TestSelectRender (0.00s)
        Error Trace:    select_test.go:64
        Error:          Not equal:
                        expected: "? Pick your word: buz "
                        received: "? Pick your word: buz"
        Messages:       Test Select answer output
```

AlecAivazis:
LGTM - thanks again @coryb!

PR #45, opened 2017-04-25
add optional Help field for prompts

coryb:
I added `tests/help.go` so you can quickly play with it to see how it works, let me know what you think.

### DIFF
--- a/core/template.go
+++ b/core/template.go
@@ -9,6 +9,11 @@ import (
 
 var DisableColor = false
 
+var (
+	HelpInputRune = '?'
+	HelpIcon      = "ⓘ"
+)
+
 var TemplateFuncs = map[string]interface{}{
 	// Templates with Color formatting. See Documentation: https://github.com/mgutz/ansi#style-format
 	"color": func(color string) string {
@@ -16,6 +21,12 @@ var TemplateFuncs = map[string]interface{}{
 			return ""
 		}
 		return ansi.ColorCode(color)
+	},
+	"HelpInputRune": func() string {
+		return string(HelpInputRune)
+	},
+	"HelpIcon": func() string {
+		return HelpIcon
 	},
 }
 
--- a/confirm.go
+++ b/confirm.go
@@ -14,23 +14,27 @@ type Confirm struct {
 	core.Renderer
 	Message string
 	Default bool
+	Help    string
 }
 
 // data available to the templates when processing
 type ConfirmTemplateData struct {
 	Confirm
-	Answer string
-	Error  *error
+	Answer   string
+	Error    *error
+	ShowHelp bool
 }
 
 // Templates with Color formatting. See Documentation: https://github.com/mgutz/ansi#style-format
 var ConfirmQuestionTemplate = `
+{{- if .ShowHelp }}{{color "cyan"}}{{ .Help }}{{color "reset"}}{{"\n"}}{{end}}
 {{- if .Error }}` + ErrorTemplate + `{{end}}
 {{- color "green+hb"}}? {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }} {{color "reset"}}
 {{- if .Answer}}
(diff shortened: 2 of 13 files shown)


coryb:
Wondering if we should make the Help prompts a bit fancier with an information emoji `ℹ️ ` or unicode infomation circle `ⓘ`.

AlecAivazis:
Hey @coryb - you're on fire! So I showed this around to some co-workers and everyone really liked it! I'm still trying to convince myself completely that `?` is the right sequence and not something more unique like `--help`.

In regards to the fancier help text, emojis don't look very good against terminals with custom colors since the color scheme is sort of out of control of the user. What we might do is add these characters and colors as exported values that could be changed before `Ask` is called which would allow for a user to easily customize the prompts to fit their needs without having to define a whole new template

coryb:
Yeah, I can see the emoji would look too weird on the console, but `ⓘ` is just a unicode sequence which can have custom colors applied.  I think I would vote for the unicode version as default.

I have been leaning against having a bunch of icon variables for users to customize, but perhaps it makes sense.

With out icon variables users could still do:
```go
survey.InputQuestionTemplate = strings.Replace(survey.InputQuestionTemplate, "ⓘ", "!")
```
vs
```go
survey.HelpInfoIcon = "!"
```
Which would change our template from this:
```go
var InputQuestionTemplate = `
{{- if .ShowHelp }}{{- color "cyan"}}ⓘ {{ .Help }}{{color "reset"}}{{"\n"}}{{end}}
...`
```
to this:
```go
var InputQuestionTemplate = `
{{- if .ShowHelp }}{{- color "cyan"}}` + HelpInfoIcon + `{{ .Help }}{{color "reset"}}{{"\n"}}{{end}}
...`
```

So I think it might makes sense to make the templates a bit more ugly with by having several icon variables.

As for the `?` vs something like `--help`.  I think `?` is pretty simple and straight forward, as a user I would be annoyed by having to type something more complex like `--help`.  Also for the Select and MultiSelect help we really need it to be a single rune so that the OnChange logic can easily react to it (vs collecting a bunch of runes in a buffer and maybe reacting eventually to a help sequence).

We could make a `HelpRune` that users could customize, `?` could be default but users to use `h` or `^` or something.
-Cory

coryb:
Oh, and of course the `?` help trigger would prevent prompts from accepting `?` as an answer only if the `Help` field is populated, if there is no `Help` set, `?` will be treated literally.  Perhaps it makes the most sense to expose a `HelpRune` to work around the edge case where someone wants `?` as a valid answer.

AlecAivazis:
I like the unicode character - assuming it lines up with the rest of the text, let's go with that. 

Also, I was also sort of against making the package super customizable at the onset because I didn't want to bloat the API, but I think I also have started to see the benefit in being able to tweak the survey to fit the user's needs and this does so with very small overhead 👍

AlecAivazis:
I'm also in favor of having the `HelpRune` be customizable by the user. I can see someone wanting to have `?` be a valid response, but this way we make them decide on what a reasonable alternative is

core/template.go
```diff
@@ -17,6 +22,12 @@ var TemplateFuncs = map[string]interface{}{
 		}
 		return ansi.ColorCode(color)
 	},
+	"HelpInputRune": func() string {
```
coryb: I have addd these helper functions to the templates to allow for dynamic prompt configuration.  The way I had originally envisioned like:
```go
var PromptTemplate := `...` + HelpIcon + `...`
```
will not work because `PromptTemplate` is not redefined when user changes `HelpIcon`.  In order to get the template make `HelpIcon` and `HelpInputRune` dynamic I just added simple template wrappers to return the dynamic value.

coryb:
I added code for allowing customizations to `HelpIcon` and `HelpInputRune`, I will add support for other Icons in a separate PR later.

AlecAivazis:
Thanks @coryb, merging now. I'm going to hold off on cutting a release with this until I can get some time tonight to add documentation to the readme

PR #47, opened 2017-04-26
make prompt icons more easily configurable

coryb:
@AlecAivazis, here is the follow up on #45 to make all the icons more easily customizable by users.

### DIFF
--- a/core/template.go
+++ b/core/template.go
@@ -11,7 +11,16 @@ var DisableColor = false
 
 var (
 	HelpInputRune = '?'
-	HelpIcon      = "ⓘ"
+
+	ErrorIcon    = "✘"
+	HelpIcon     = "ⓘ"
+	QuestionIcon = "?"
+
+	MarkedOptionIcon   = "◉"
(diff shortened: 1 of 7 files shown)


core/template.go
```diff
@@ -11,7 +11,16 @@ var DisableColor = false
...
-	HelpIcon      = "ⓘ"
+
+	ErrorIcon    = "✘"
+	HelpIcon     = "ⓘ"
+	QuestionIcon = "?"
+
+	MarkedOptionIcon   = "◉"
+	UnmarkedOptionIcon = "◯"
+
+	MultiSelectFocusIcon = "❯"
```
AlecAivazis: Should these be the same for consistency?

coryb: Yeah, I was thinking that, consistency would probably be a good idea.  Which one wins?

AlecAivazis: imo `MultiSelectFocusIcon` is a pretty clear winner

coryb: Cool, my preference also, PR incoming :)

PR #48, opened 2017-04-27
use same selection focus icon for Select and MultiSelect prompts

coryb:
@AlecAivazis merging icons as per discussion in #47

### DIFF
--- a/core/template.go
+++ b/core/template.go
@@ -19,8 +19,7 @@ var (
 	MarkedOptionIcon   = "◉"
 	UnmarkedOptionIcon = "◯"
 
-	MultiSelectFocusIcon = "❯"
-	SelectFocusIcon      = ">"
+	SelectFocusIcon = "❯"
 )
 
 var TemplateFuncs = map[string]interface{}{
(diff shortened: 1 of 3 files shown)


AlecAivazis:
So what do you think about still having two different icons with the same value? that way if the user wants to define them as different things then they can

coryb:
Sure, if you want.  Otherwise user could always just set the prompt between invocations of `Ask` to have the different icons.  Hard to say, I would expect the common case to be using the same icon (maybe you have a different experience with that?), and so asking user to write a few extra lines of code for an uncommon case does not seem to bad.   So I guess my vote is to have common icon, but your vote weighs more :) so up to you.

AlecAivazis:
Yea I think you're probably right that 99.9% of the time people will want it to be the same so i'm fine with that for now. 

Also, you bring up an interesting point, since you've been doing so much work on this library in the past month, you probably know the codebase as well as I do - how would you feel about becoming an official contributor? I would love to have someone look over these PRs as things go on, and since I plan to maintain this for the foreseeable future, it would be great to have an extra set of hands helping around. No pressure if you're not interested, just figured I would ask.

coryb:
Hey, sure would love to help out when I can.  Thanks, I promise not to randomly commit gibberish when you aren't looking :)

AlecAivazis:
Great! I'll add you now, also FWIW i spend a good amount of time in the golang slack channel if you ever want to get ahold of me and github isn't the right venue

Issue #56, opened 2017-05-10
Input prompt on Answer renders wrong if answer is ""


coryb:
Creating issue for problem I saw working on #55.  The template has:
```
{{- if .Answer}}
  {{- color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
```
So if the default answer is `""` then it evaluates to false and the template shows the prompt again.  We probbaly need a `.ShowAnswer` boolean flag for the template to know when we need to show an answer for a value that will be evaluated to false by `text/template`.

PR #57, opened 2017-05-11
fixed input cleanup when default is empty string

AlecAivazis:
This PR fixes a bug reported in #56 which was causing an issue when `Inputs` were cleaning up without a default value.

### DIFF
--- a/input.go
+++ b/input.go
@@ -18,8 +18,9 @@ type Input struct {
 // data available to the templates when processing
 type InputTemplateData struct {
 	Input
-	Answer   string
-	ShowHelp bool
+	Answer     string
+	ShowAnswer bool
+	ShowHelp   bool
 }
 
 // Templates with Color formatting. See Documentation: https://github.com/mgutz/ansi#style-format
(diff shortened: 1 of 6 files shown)


coryb:
looks good.  We should probably make the same change to Select and MultiSelect templates in case a user wants to add "empty" option for some reason?

AlecAivazis:
Yea, I was just putting that together. The tests seem to run fine on my end

AlecAivazis:
Thanks @coryb!

AlecAivazis:
I should have a chance to test the windows stuff tomorrow so i'll wait to tag this and include it with the full windows support (probably worth a minor version bump since readline is gone)

coryb:
Okay, cool.  We should be able to add `.Help` to Password after we merge the Readline PR as well.   Not sure if you want to wait for that before releasing.  I am also planning on updating the autoplay PR immediately after the Readline PR.   I updated some autoplay tests on my system and ran them in a loop or a few thousand times with no errors, so I think the random EOF's we were seeing from Readline are gone as well.

AlecAivazis:
Amazing! Okay, i'm going to see if I can figure something out to get that PR in tonight.

PR #59, opened 2017-05-11
add Help to Password prompt

coryb:
Change should look pretty familar, mostly just copy/paste from Input to allow for Help.  Had to change the template data type to allow for ShowHelp field.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,7 +181,7 @@ validators include:
 
 ## Help Text
 
-All of the prompts (except `Password`) have a `Help` field which can be defined to provide more information to your users:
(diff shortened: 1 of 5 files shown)


coryb:
doh, `tests/password.go` not working, debugging now.

coryb:
okay, `test/password.go` working again.  I forgot that `rr.ReadLine` returns `[]rune` rather than `string` ... the problems of using `interface{}` types :)

AlecAivazis:
Works like a charm - thanks! And that's why we have the tests :P They have definitely caught me on a number of occassions doing similar things.

PR #60, opened 2017-05-11
adding autoplay tests to verify the prompt output is not changing

coryb:
Replacing PR from #51 that seemed crazy, maybe bad merge.


coryb:
Okay, started a new branch from current master then merged the changes from the first PR, so now the changes seem sane.  Still passes Travis, so that is good.

AlecAivazis:
Awesome, taking a look now - should be able to get this in before I go to sleep tonight.

AlecAivazis:
Hm not sure why the current build is failing, I pushed a small commit and now travis is complaining 😞

coryb:
Not sure, I will take a look.

coryb:
So far no idea on why it would fail that way.  I reran the failed build from last night and it was successful on retry.  I am running the entire regression in a loop on my system and so far no errors, I will do the same on a linux box when I get in to work.

coryb:
Don't have a concrete answer.  I ran it over a thousand times today on my OSX laptop and several thousand times on Linux.  It did fail a few times.  There is a race condition with pty in the echoed output when typing in an answer to a prompt.  The ultimate answer is to probably add some artificial millisecond pauses after input to wait for echoed output.  So we will probably see a few errors on travis from this, but retries should fix it, and I would expect it to fail very infrequently because of this.  I will keep an eye on it and see if I can come up with some fixes to `autoplay` to better handle IO race conditions.

Issue #62, opened 2017-05-14
failed validation does not cleanup prompt on answer


coryb:
I noticed while testing that if you fail a validation multiple times it will stack up the error messages.  So this test code:
```
    answer := ""
    survey.AskOne(
        &survey.Input{Message: "Enter something:"},
        &answer,
        survey.Required,
    )
    fmt.Printf("response string: %s\n", answer)
```

Will produce this if you just type `Enter` a few times:
```
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
✘ Sorry, your reply was invalid: Value is required
? Enter something: something
response string: something
```
I am not sure if this is desired behavior or not.

Since the `renderer` does know now about printing the error message it is unable to include those lines in the `lineCount` used when rewriting the prompt.  If we wanted to fix this I think we need to somehow make the prompt renderer render the error message upon validation, but not sure what changes would be required to make this happen.

PR #63, opened 2017-05-15
cleanup error message on repeated validation failures

coryb:
This is a fix for #62 to allow prompts to clean up previous error messages on repeated failed validations.

It basically adds an `Error(error) error` function to the `Prompt` interface which is implemented via the `renderer` common base composed class.


AlecAivazis:
Thanks for submitting yet another PR @coryb, just ran through the tests and everything passed 👍

Issue #66, opened 2017-05-23
Add support for map[string]interface{} as a result


jdoklovic:
Currently you can pass a struct with fields to gather answers, however this is not very dynamic and requires the fields to be known up front.

It would be great to be able to pass map[string]interface{} as the thing to populate answers, however currently it fails due to type casting...
`reflect.Set: value of type string is not assignable to type map[string]interface {}`

AlecAivazis:
Closing this alongside #67

PR #67, opened 2017-05-23
answer maps work, needs tests

jdoklovic:
quick fix for #66 
I didn't have time to figure out how to add autoplay tests... please help

### DIFF
--- a/examples/map.go
+++ b/examples/map.go
@@ -26,18 +26,15 @@ var simpleQs = []*survey.Question{
 }
 
 func main() {
-	answers := struct {
-		Name  string
-		Color string
-	}{}
+	ansmap := make(map[string]string)
 
 	// ask the question
-	err := survey.Ask(simpleQs, &answers)
+	err := survey.Ask(simpleQs, &ansmap)
 
 	if err != nil {
 		fmt.Println(err.Error())
 		return
 	}
 	// print the answers
-	fmt.Printf("%s chose %s.\n", answers.Name, answers.Color)
+	fmt.Printf("%s chose %s.\n", ansmap["name"], ansmap["color"])
 }
--- a/examples/simple.go
+++ b/examples/simple.go
@@ -26,17 +26,20 @@ var simpleQs = []*survey.Question{
 }
 
 func main() {
-	answers := struct {
-		Name  string
-		Color string
-	}{}
+	//	answers := struct {
+	//		Name  string
+	//		Color string
+	//	}{}
+
+	ansmap := make(map[string]interface{})
+
 	// ask the question
-	err := survey.Ask(simpleQs, &answers)
+	err := survey.Ask(simpleQs, &ansmap)
 
 	if err != nil {
 		fmt.Println(err.Error())
 		return
 	}
 	// print the answers
-	fmt.Printf("%s chose %s.\n", answers.Name, answers.Color)
+	fmt.Printf("%s chose %s.\n", ansmap["name"], ansmap["color"])
 }
--- a/core/write.go
+++ b/core/write.go
@@ -38,8 +38,12 @@ func WriteAnswer(t interface{}, name string, v interface{}) (err error) {
 		// copy the value over to the field
 		return copy(elem.Field(fieldIndex), value)
 	case reflect.Map:
-		mt := *t.(*map[string]interface{})
-		mt[name] = value
+		mapType := reflect.TypeOf(t).Elem()
+		if mapType.Key().Kind() != reflect.String || mapType.Elem().Kind() != reflect.String {
+			return errors.New("answer maps must be of type map[string]string")
+		}
+		mt := *t.(*map[string]string)
+		mt[name] = value.Interface().(string)
 		return nil
 	}
 	// otherwise just copy the value to the target
--- a/core/write.go
+++ b/core/write.go
@@ -39,11 +39,11 @@ func WriteAnswer(t interface{}, name string, v interface{}) (err error) {
 		return copy(elem.Field(fieldIndex), value)
 	case reflect.Map:
 		mapType := reflect.TypeOf(t).Elem()
-		if mapType.Key().Kind() != reflect.String || mapType.Elem().Kind() != reflect.String {
-			return errors.New("answer maps must be of type map[string]string")
+		if mapType.Key().Kind() != reflect.String || mapType.Elem().Kind() != reflect.Interface {
+			return errors.New("answer maps must be of type map[string]interface")
 		}
-		mt := *t.(*map[string]string)
-		mt[name] = value.Interface().(string)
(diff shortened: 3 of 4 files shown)


AlecAivazis:
Hey @jdoklovic sorry for not responding to your two issues earlier, and thanks for getting this PR out so quickly! adding autoplay tests is a bit involved at the moment but @coryb or I can do that once this PR is in. 

Would you mind adding a unit test for writing the answer as a map? The rest of the tests for `WriteAnswer` are here: https://github.com/AlecAivazis/survey/blob/master/core/write_test.go

examples/simple.go
```diff
@@ -26,17 +26,20 @@ var simpleQs = []*survey.Question{
 }
 
 func main() {
-	answers := struct {
-		Name  string
-		Color string
-	}{}
+	//	answers := struct {
```
AlecAivazis: Can we leave this file as is? I much prefer showing the struct pattern as the default one since it avoids unnecessary type casting and lends to cleaner code.

jdoklovic: sure. I'll just add another example.

jdoklovic:
I'll add a unit test today

jdoklovic:
@AlecAivazis I added some tests and fixed up the map type as it's required to be map[string]string

examples/map.go
```diff
@@ -0,0 +1,40 @@
...
+		Name: "color",
+		Prompt: &survey.Select{
+			Message: "Choose a color:",
+			Options: []string{"red", "blue", "green"},
+		},
+		Validate: survey.Required,
+	},
+}
+
+func main() {
```
AlecAivazis: So i've been thinking about this more and more and I'm not convinced that this is a good api to add to `survey` since it tends to a lot of type assertions. Do you mind adding an example that cannot be accomplished with a `struct`? My gut says that if the want is for a dynamic set of questions, that can be accomplished with #5. I'm definitely interested in exploring this, just want to make sure we make very deligent additions to the API, removing them is hard.

AlecAivazis:
Thanks for adding the tests @jdoklovic - I'm a bit dubious about the `map[string]string` since `Confirm` returns a `bool`. I'll do some testing when I have some down time to run the tests and make sure thigns work with the map.

jdoklovic:
@AlecAivazis you were right... updated to be map[string]interface{}

examples/map.go
jdoklovic: 2 things...
1. I'm writing a templating CLI where a json file can be used to setup the prompts by anyone using the cli. I need to turn these into questions and get the answers so i need to build the answers container dynamically

2. I'm almost finished with another PR that you're probably not going to like, but it adds a bunch of type assertions to the copy method inside of write that can figure out the type of field on a struct and convert string values to it. e.g. you can use a struct with fields that are bool, int, float, so on and the answers will be converted to those types instead of structs only allowing string fields.

3. In the same PR, I'm adding a Converter func that's a lot like the Validator, but can be used to apply type conversion before setting the answers.

AlecAivazis: Yea tbh, this makes me really nervous. I'm also not sure what you mean by structs only allowing string fields - there are many examples of non-string fields in the codebase.  

I think we need to explore what the intended use case is and try to figure out what the simplest way for the user to specify that is. I think i'm much more in favor of adding `When` to `survey.Question` which would allow for this sort of dynamic logic withou a lot of complication.

jdoklovic: Here's what I mean... if you have this test:
```go
package survey

import (
	"io"
	"io/ioutil"
	"os"
	"testing"

	"github.com/AlecAivazis/survey/core"
	"github.com/stretchr/testify/assert"
)

func init() {
	// disable color output for all prompts to simplify testing
	core.DisableColor = true
}

func TestStructInt(t *testing.T) {
	q := []*Question{
		{
			Name: "age",
			Prompt: &Input{
				Message: "What is your age?",
			},
			Validate: Required,
		},
	}

	in, _ := ioutil.TempFile("", "")
	defer in.Close()

	os.Stdin = in

	io.WriteString(in, "21\n")
	in.Seek(0, os.SEEK_SET)

	ans := struct{ Age int64 }{}

	err := Ask(q, &ans)

	assert.Nil(t, err)
	assert.Equal(t, int64(21), ans.Age)

}
```
this is what you get when you run it:
```
=== RUN   TestStructInt
[0G? What is your age? 21
[1F[0G? What is your age? 21
--- FAIL: TestStructInt (0.00s)
	assertions.go:241: 
                        
	Error Trace:	struct_int_test.go:41
		
	Error:		Expected nil, but got: &errors.errorString{s:"reflect.Set: value of type string is not assignable to type int64"}
		
	assertions.go:241: 
                        
	Error Trace:	struct_int_test.go:42
		
	Error:		Not equal: 21 (expected)
			        != 0 (actual)
		
FAIL
exit status 1
FAIL	command-line-arguments	0.003s
```

jdoklovic: This is going to make you even more uncomfortable:
https://github.com/AlecAivazis/survey/compare/master...brainicorn:handle-field-types

If you want these changes, I can make a PR, if not, I can work around it by using my own impl of Ask and WriteAnswer

AlecAivazis: Sorry to ask this, but i'm confused what the intended use case is for doing this sort of arbitrary type conversion from a string? Seems like if the user wants a specific type out, they should just use the prompt that collects that type, or do the conversion from the provided type to the intended one themselves.

AlecAivazis: I'd also like to avoid the additional argument to `AskOne` as that would introduce an API breaking change. I really appreciate all the effort and hard work you've put into these changes, but I dont think I want to incorporate them just yet. I'll need to do a lot of thinking about the "role" of survey and what sorts of things to push as use-case specific layers on top.

AlecAivazis:
So after thinking about your problem of creating a survey out of a json file, it occured to me that you could build up your map by iterating over the entries in your json file and calling `AskOne` for each. This feels logically equivalent to building up the list of questions first and then calling `Ask` but maybe i'm missing something.

examples/map.go
jdoklovic: I assume you're talking about using AskOne and passing in the type I want out. This definitely works, but if I have 5+ prompts with various types and need to eventually get them into a map[string]interface{} to pass them as data to a go template, that's a lot more code on the caller side then simply passing slice of questions to Ask and getting back the proper types.
Doing it one by one is a bit of a pain.

AlecAivazis: Alright I was skeptical at first you make a good point. You've convinced me supporting map[string] interface{} does make sense since it's relatively straight forward but id like to avoid the string assertion PR since that seems to be a bigger change that I want to think about a bit deeper. I'll go through this PR closely today and will let u know if there's anything else that needs changing.

Issue #71, opened 2017-06-28
MultiSelect does not support cycling through the list


ahmetb:
If I hit <kbd>Up</kbd> key at the beginning of the list, it should go to the item at the end of the list.

Similarly, if I hit <kbd>Down</kbd> at the end of the list, it should cycle back to the item at the beginning of the list.

This behavior is supported by Inquirer.js.

canghai908:
I also have this questions

AlecAivazis:
Thanks again for reporting this @ahmetb!

@ahmetb @canghai908 - I decided to spend a little time today to address this have fixed it in a branch. I'm going to look into fixing #70 as well before I cut a new release but hopefully that won't take too long. I'll keep this issue open until the branch is merged and let u guys know when u can expect to see this in your projects.

AlecAivazis:
@ahmetb @canghai908 you should be able to see the current selects cycling through the pages when you navigate past the bounds - lemme know if you run into any more issues.

Issue #70, opened 2017-06-28
MultiSelect fails to render a long list


ahmetb:
Below you can see a picture of a long list passed to MultiSelect. Issues:

- default view is the end of the list, even though the cursor is by default at item 0.
- keeps showing only the end of list no matter where the cursor is moved to.
    - never shows the earlier items. I tried bringing cursor all the way to the end and tried to bring it above Item 102. It never showed Item 101 or things before it.

![image](https://user-images.githubusercontent.com/159209/27660092-b5dfd0b8-5c0a-11e7-8262-119eacf00f6a.png)

MultiSelect should:
- adapt to the dynamic terminal height
- slide the options window displayed along with the cursor

ahmetb:
@AlecAivazis I recommend looking at Inquirer.js `examples/long-list.js`. It does not use all of the screen height to list the options, but offers a sliding window in e.g. 7 rows:

![](http://g.recordit.co/2rgsxcNvkN.gif)

It could be offered as an option, it's not really necessary to me at least.

AlecAivazis:
@ahmetb thanks for the screenshots and pointing me to the example in inquirer. I'm not sure about the specific details of their implementation but my thinking is this - add a pageSize option to Select and MultiSelect and then show the minimum of that number and the terminal height. that way even if the developer specified a large number because the list is long, but the user has a terminal that is short, there isn't any potential for lost information (say if they turned a scrollback off

ahmetb:
Yeah that's tricky. I'd play with Inquirer to see how they handle this case if you want to get closer to their implementation.

AlecAivazis:
@ahmetb as of the `v1.2.0` tag, the selects and multi-selects should be paginated with a default value of 7 - lemme know if you run into any more issues

Issue #61, opened 2017-05-14
usability issues for Select and MultiSelect with many options


coryb:
As I run through my projects and replace prompts with `survey` I found a few minor usability issues with `Select` and `MultiSelect`.  Specifically if they have many options it can be too much to present all at once.   So here is a dumb example:
```
? Pick a number:
  1
  2
  3
  4
❯ 5
  6
  7
  8
  9
  10
```
I would like to add a property to limit the view, perhaps `ViewSize` so we can focus on just a few options.  So setting `ViewSize: 5` we would see something like:
```
? Pick a number:
  3
  4
❯ 5
  6
  7
```

But we would probably need some markers to indicate there is more Up or Down, so maybe something more like:
```
? Pick a number:
▲ 3
  4
❯ 5
  6
▼ 7
```

The next issue is that I would like to enable text focusing for long lists.  If you have a `Select` list of all the US States, you currently would have to hit the Down arrow 49 times to get to Wyoming.  Ideally we allow users to just type in `W` and jump the focus to the closest match starting with the letter.

AlecAivazis:
Good point @coryb. I think the up/down indicators are a good solution, but I'll keep thinking about different ways to solve this in case i think of something different. Also, I like the idea of filtering on a select/multi-select - does the text show up while the user types?

I wonder if having a long list like the examples you showed is reason to look into a multi-column prompt...

AlecAivazis:
Something tells me we would be able to figure out how big to make the pages by looking at the cursors current location and comparing it to the terminal height. Not sure what the windows calls look like for that just yet, but I suspect it's doable.

coryb:
The text does not show up while typing in select/multi-select, `echo` is disabled in the terminal so that we can interpret up/down arrows etc.

My feeling is that multi-column prompts would probably be pretty confusing to users, at least I don't recall ever having seen a multi-column prompt on a console.  It might be an interesting option to play with, but I would want that behavior optional.

I think a reasonable default for the view size would be the terminal hight (when terminal height is smaller then options size), but I generally want the view size to be much smaller, probably closer to 5 lines typically.

AlecAivazis:
Currently, it is true that `echo` is disabled on those prompts but if we end up allowing the user to jump to particular options, we probably want to give the user a way to see what they're typing. It doesn't seem likely that the options will always be in alphabetical order so if the user wants to search by one character, they will probably want to enter a second.

[...]

PR #72, opened 2017-07-04
Feat/paginated select

AlecAivazis:
This PR adds a few major improvements to the package in order to address a few issues that were recently reported as well as some things I ran into while writing those fixes. The important ones are:

1. Select and MultiSelect cycle through the top and bottom when the user presses the appropriate key [#71]
1. Select and MultiSelect are paginated [#70, #61] 
    1. default page size is 7
    1. both prompts have a `PageSize` field which allows for independent page sizes for prompts
    1. page size can be changed by setting `survey.PageSize` to another value
1. Improved the readline behavior to allow for the cursor to be moved with the arrows keys and insert text anywhere in the line


coryb:
Hey @AlecAivazis, I will try to dig into this PR, there is a lot of changes to review here.  Not sure why the test failed also.

AlecAivazis:
Thanks @coryb. Sorry about the high number of changes - I wanted to have it be 3 separate PRs but had very limited access over vacation and one thing led to another... I'm planning on investigating the failing test in the next few days

multiselect.go
```diff
@@ -28,9 +29,10 @@ type MultiSelectTemplateData struct {
 	MultiSelect
 	Answer        string
 	ShowAnswer    bool
-	Checked       map[int]bool
+	Checked       map[string]bool
 	SelectedIndex int
 	ShowHelp      bool
+	Choices       []string
```
coryb: Wondering if we should change the name of this, maybe `Page`, `PageOptions` or something like that?  `Choices` and `Options` are indistinguishable to me.

survey_test.go
```diff
@@ -42,3 +42,97 @@ func TestAsk_returnsErrorIfTargetIsNil(t *testing.T) {
...
+	choices := []string{"choice1", "choice2", "choice3"}
+	// a page bigger than the total number
+	pageSize := 4
+	// the current selection
+	sel := 2
+
+	// compute the page info
+	choices, idx := paginate(pageSize, choices, sel)
+
+	if choices[0] != "choice1" && choices[1] != "choice2" && choices[2] != "choice3" {
```
coryb: using testify like with other tests, I think this all simplifies to:

```
page, idx := paginate(pageSize, choices, sel)
assert.Equal(t, choices, page)
assert.Equal(t, 2, idx)
```

survey_test.go
```diff
@@ -42,3 +42,97 @@ func TestAsk_returnsErrorIfTargetIsNil(t *testing.T) {
...
+	pageSize := 4
+
+	// test the second item
+	sel := 2
+
+	// compute the page info
+	choices, idx := paginate(pageSize, choices, sel)
+
+	// we should see the first three options
+	if choices[0] != "choice1" && choices[1] != "choice2" && choices[2] != "choice3" && choices[3] != "choice3" {
```
coryb: again here simplify with testity, also fix logical error (need `||` vs `&&`)
```
page, idx := paginate(pageSize, choices, sel)
assert.Equal(t, choices[0:4], page)
assert.Equal(t, 2, idx)
```

survey_test.go
```diff
@@ -42,3 +42,97 @@ func TestAsk_returnsErrorIfTargetIsNil(t *testing.T) {
...
+		t.Error("Did not recieve the correct index")
+	}
+
+}
+
+func TestPagination_firstHalf(t *testing.T) {
+	// the choices for the test
+	choices := []string{"choice1", "choice2", "choice3", "choice4", "choice5", "choice6"}
+
+	// section the choices into groups of 3
```
coryb: comment does not match value, not sure which you intended

tests/select.go
```diff
@@ -39,6 +39,18 @@ var goodTable = []TestUtil.TestTableEntry{
 			Options: []string{"red", "blue"},
 		}, &answer,
 	},
+	{
```
coryb: you will need to regenerate the autoplay scripts since you changed the test

tests/selectThenInput.go
```diff
@@ -16,13 +16,6 @@ var simpleQs = []*survey.Question{
 		},
 		Validate: survey.Required,
 	},
-	{
```
coryb: will need to update autoplay tests since you removed tests

select.go
```diff
@@ -25,6 +26,7 @@ type Select struct {
 // the data available to the templates when processing
 type SelectTemplateData struct {
 	Select
+	Choices       []string
```
coryb: again here, I think we need to pick a different name, `Page` etc.

coryb:
Hey, I am going to be out of town from Saturday thru Thursday with no internet access, so I am just going to approve this PR now.  Hopefully you can still get updates for the autoplay scripts.  The travis autoplay error is still confusing to me, it works on my osx laptop.  If they are still failing after you update for your changes then I will take a look, might be linux specific.

AlecAivazis:
Awesome, thanks for looking at this @coryb  - I updated the autoplay tests and there seems to be an error in travis with a double carriage return on linux 😞

coryb:
Okay, I confirmed the tests are working on OSX.  I think the issue is a timing issue on linux for some reason.  I can reproduce via a docker container, not sure why it so random.  
Anyway, to get travis working, it is probably best for now to just comment out the `run autoplay-tests` in the `.travis.yml` and we can try to get it more stable in the future.

AlecAivazis:
yea I think you're right, i'm going to try to re-run the tests and if that doesn't work i'll comment out the line in the travis config - thanks for verifying that the autoplay tests pass for you!

PR #74, opened 2017-07-13
allow for settable interfaces

coryb:
This is a little abstract, but it will allow for the users to be able to control how assignment is done to custom types via the survey.Ask routines, something along the lines of:

```
survey.AskOne(
    &survey.Input{
        Message: "Enter something:",
   },
   &myCustomType,
   nil,
)
```

This patch allows me to implement: `Set(value interface{}) error` or `SetField(name string, value interface{}) error` on my custom type to perform the assignments.  

I was working on refactoring a large project that uses `survey` and ran into problems as a `string` was assigned to `MyCustomType` via the `core.copy` routine.

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -10,7 +10,24 @@ import (
 // the tag used to denote the name of the question
 const tagName = "survey"
 
+// add a few interfaces so users can configure how the prompt values are set
+type settable interface {
+	Set(value interface{}) error
+}
+
+type fieldsettable interface {
+	SetField(field string, value interface{}) error
+}
+
 func WriteAnswer(t interface{}, name string, v interface{}) (err error) {
+
+	if s, ok := t.(settable); ok {
+		return s.Set(v)
+	}
+	if fs, ok := t.(fieldsettable); ok {
+		return fs.SetField(name, v)
+	}
+
 	// the target to write to
 	target := reflect.ValueOf(t)
 	// the value to write from
--- a/core/write.go
+++ b/core/write.go
@@ -12,20 +12,20 @@ const tagName = "survey"
 
 // add a few interfaces so users can configure how the prompt values are set
 type settable interface {
-	Set(value interface{}) error
+	WriteAnswer(value interface{}) error
 }
 
 type fieldsettable interface {
-	SetField(field string, value interface{}) error
+	WriteAnswerField(field string, value interface{}) error
 }
 
 func WriteAnswer(t interface{}, name string, v interface{}) (err error) {
 
 	if s, ok := t.(settable); ok {
-		return s.Set(v)
+		return s.WriteAnswer(v)
 	}
 	if fs, ok := t.(fieldsettable); ok {
-		return fs.SetField(name, v)
+		return fs.WriteAnswerField(name, v)
 	}
 
 	// the target to write to
--- a/README.md
+++ b/README.md
@@ -194,6 +194,7 @@ survey.AskOne(
     &myval,
     nil,
 )
+// myval.value should be populated with the prompt result now.
 ```
 
 If you want to capture the name associated with the prompt you can use this form:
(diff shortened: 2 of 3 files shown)


core/write.go
```diff
@@ -10,7 +10,24 @@ import (
 // the tag used to denote the name of the question
 const tagName = "survey"
 
+// add a few interfaces so users can configure how the prompt values are set
+type settable interface {
+	Set(value interface{}) error
```
coryb: I wonder if this should be something like `SetAnswer` instead.  The kingpin library has a similar interface `Set(string)` which would potentially conflict with this so a type could not be both compatible with kingpin and survey in this case.

coryb: Of course we could use the same name as your function so require a `WriteAnswer(interface{}) error`

coryb:
I have modified the code to use `WriteAnswer(value interface{} error` and `WriteAnswerField(name string, value interface{}) error` to make the interface more clear when someone is providing these routines.

AlecAivazis:
Hey @coryb,

This reminds me of a PR that @jdoklovic showed me awhile ago where he had added the necessary changes to `Write` in order to support casts from strings to the basic types. So the following would work, for example:

```golang
var foo int 

survey.AskOne(
    &survey.Input{
        Message: "Enter something:",
   },
   &foo,
   nil,
)
```

Is this something you think would be good to add? I wanted to wait until a second person had requested it and the code is already there and tested. So if we want to support writing to variables other than the explicit types returned by the prompts, I can see an argument to do both of these in a single release - thoughts?

coryb:
We could add it, I already have the conversion code in another project I maintain. This particular code is for converting command line arguments into primitive types, so the requirements are very similar:
https://github.com/coryb/figtree/blob/master/convert.go#L9  This code I would imagine would replace your `copy` function that uses reflection since the `v interface{}` values is almost always a `string`

That said, by merging in this PR users can implement the int-assignment on their own with a reasonably simple wrapper:
```
type intPrompt struct {
    value *int
}
func (ip *intPrompt) WriteAnswer(value interface{}) (err error) {
    var tmp int64
    if str, ok := value.(string); ok {
        tmp, err = strconv.ParseInt(str, 10, 32)
    } else {
        err = fmt.Errorf("Invalid answer of type %T", value)
    }
    if err != nil {
        return err
    }
    *ip.value = int(tmp)
    return nil
}
```

Which is used like this:
```
var foo int
survey.Ask
    &survey.Input{
        Message: "Enter something:",
    },
    &intPrompt{&foo},
    nil,
)
```

AlecAivazis:
Hm do you have this code in place because survey doesn't support different types? Here is the PR by @jdoklovic where he seems to have solved the big cases: https://github.com/AlecAivazis/survey/compare/master...brainicorn:handle-field-types

This is starting to seem way more common than I originally thought so I'm definitely starting to be more open to the idea.

I think your point about being able to support his use case with this interface really speaks to its power, but I'd like to save the user some effort if we can. I'm going to review this during lunch today and revive that other PR

coryb:
I wanted the interface so I can apply assignment logic to my custom types since they are not primitives.  So even if you merge in the other PR to deal with primitives I still think it is a good idea to allow for assignment via an interface to allow for users to have custom types.   To be more explicit I have a type like this:
```
type StringOption struct {
    Value string
    Source string
}
```
It is a bit convoluted but I have a need to track how my StringOption is populated (ie goal is to track if it was loaded from a config file, assigned via command line options (using kingpin) or populated via a prompt in survey.

So basically with survey I want to do something like:
```
strOpt := StringOption{
    Value: "default",
    Source: "default",
}
survey.AskOne(
    &survey.Input{
        Message: "Enter something:",
   },
   &strOpt,
   nil,
)
```

AlecAivazis:
Yea I definitely think these two PRs go together really well for the exact reasons you bring up - I think the combination provides a very nice solution to allowing for the user to specify the types they want and having survey do something that makes sense by default and is easily adaptable.

AlecAivazis:
@coryb - do you mind adding something in the README detailing this interface? Probably below all of the prompt examples.

coryb:
Okay added some brief documentation for how to use with custom types.

AlecAivazis:
Awesome thanks. I'm going to merge this now and then cut a release

PR #75, opened 2017-07-14
Enable converting from prompt response to primitive types

AlecAivazis:
This PR (thanks to @jdoklovic) allows for easy conversion from the return type of the prompt to the specified one in the given struct. 

I have a few minor things I'd like to change before this is ready, but those should get in pretty quick since all the hard work is already done.

One thign i'd like to do is move the Converter work to a different PR, i think it can be merged in tandem with a `Transform` field that I have been thinking about allowing before values get validated.

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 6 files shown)


coryb:
Hey this looks good, I like the changes to WriteAnswer to do the implicit conversion.  I recommend removing the `Converter` argument from AskOne so we don't break that API.  I think that functionality would be addressed by allowing for an interface type on WriteAnswer arguments as covered in PR #74.

AlecAivazis:
Perfect, that was my exact thinking as well. I'll
Make that change today and hopefully we can cut a release.

core/write.go
```diff
@@ -102,6 +103,80 @@ func copy(t reflect.Value, v reflect.Value) (err error) {
...
+			castVal, casterr = strconv.ParseUint(vString, 10, 64)
+		case reflect.Float32:
+			var val64 float64
+			val64, casterr = strconv.ParseFloat(vString, 32)
+			if casterr == nil {
+				castVal = float32(val64)
+			}
+		case reflect.Float64:
+			castVal, casterr = strconv.ParseFloat(vString, 64)
+		}
```
coryb: We should probably add a default case for when someone passes in an unknown type, something like:
```
    default:
        return fmt.Errorf("Unable to convert from string to type %s", t.Kind())
```

Issue #77, opened 2017-07-17
Required Validation does not work for MultiSelect


wburningham:
The code panics if you try to use a MultiSelect question with the built in Required validator.

AlecAivazis:
Thanks for reporting this @wburningham. I was able to reproduce it and will try to get a fix out when I get back from work today.

AlecAivazis:
@wburningham I just pushed a PR with a fix for this bug. Once it's merged, i'll cut a release with the fix ASAP. Thanks again for reporting the issue!

AlecAivazis:
This should be fixed in `v1.2.2` please let me know if you're still running into it.

PR #78, opened 2017-07-18
[#77] fixed bug causing required to panic on multiselects

AlecAivazis:
This PR fixes a bug reported in #77 which caused the Required validator to panic when passed an empty list. Thanks to @wburningham for pointing this out!

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -9,7 +9,7 @@ import (
 // Required does not allow an empty value
 func Required(val interface{}) error {
 	// if the value passed in is the zero value of the appropriate type
-	if val == reflect.Zero(reflect.TypeOf(val)).Interface() {
+	if isZero(reflect.ValueOf(val)) {
 		return errors.New("Value is required")
 	}
 	return nil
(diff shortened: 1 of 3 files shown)


validate.go
```diff
@@ -71,3 +71,34 @@ func ComposeValidators(validators ...Validator) Validator {
 		return nil
 	}
 }
+
+// isZero returns true if the passed value is the zero object
+func isZero(v reflect.Value) bool {
```
coryb: Not sure this implantation covers all the caveats that make a zero object.
This is what I have used in the past:
```
reflect.DeepEqual(v.Interface(), reflect.Zero(v.Type()).Interface())
```
Basically we just create a zero value of the type passed in and verify that equals the actual value passed in.  Not saying your code wont work, clearly it does, but I am just not sure what (if any) edge cases are missing.  I am pretty confident the DeepEqual covers all the edge cases the the cost of an extra zero-allocation.  If you want to keep your implementation that is fine, just something to think about.

AlecAivazis: Even if they both work, I like your solution more since way fewer lines of code - I'll change it now and make sure the tests pass (i'm sure they will).

AlecAivazis: So I replaced the function with the following body:

```golang

// Required does not allow an empty value
func Required(val interface{}) error {
	// reflect the value
	v := reflect.ValueOf(val)
	// if the value passed in is the zero value of the appropriate type
	if reflect.DeepEqual(v.Interface(), reflect.Zero(v.Type()).Interface()) {
		return errors.New("Value is required")
	}

	return nil
}
```

And now the list test is failing, this looks a lot like what I had before, but I think this won't work for the same reason - our `MultiChoice` prompt returns an empty list of  answers, not `nil` which is what I think the `Zero()` value is of an array.

AlecAivazis:
@coryb here's a playground to see what i'm talking about: https://play.golang.org/p/XpOy2nKmMd

I was able to simplify the logic substantially though using your suggestion so thanks for that!

coryb:
Interesting apparently `var x []string` is a zero value but `x := []string{}` is not.

validate.go
```diff
@@ -71,3 +71,14 @@ func ComposeValidators(validators ...Validator) Validator {
...
+	switch v.Kind() {
+	case reflect.Slice:
```
coryb: To be pedantic, I think this should probably be:
```
case reflect.Slice, reflect.Map:
```
It turns out neither of these are considered zero values:
```
x := map[string]string{}
x := []string{}
```
So you should check the `Len` on both of them.  Slightly curiously the array version of this is fine, so `x := [4]string{}` is considered a zero value.

AlecAivazis: Nice catch, pushed the change :+1:

PR #80, opened 2017-07-19
Refactor/one write interface

AlecAivazis:
This PR merges the write interfaces into a single one with `WriteAnswer(name string, value interface{}`) in order to simplify the overall API.

Part of me thinks this interface pattern is better suited to be under a `Transform` field on the `survey.Question` but it's hard for me to know why that feels more right. I guess the interface feels a bit like unnecessary boilerplate in order to change one value to another. What do you think?

If we do opt to keep it, I'd like to condense the two interfaces into a single one as I've described above in order to minimize the number of different ways something can be done.

### DIFF
--- a/core/write_test.go
+++ b/core/write_test.go
@@ -256,6 +256,10 @@ type testFieldSettable struct {
 	Values map[string]string
 }
 
+type testTaggedStruct struct {
+	TaggedValue string `survey:"tagged"`
+}
+
 func (t *testFieldSettable) WriteAnswer(name string, value interface{}) error {
 	if t.Values == nil {
 		t.Values = map[string]string{}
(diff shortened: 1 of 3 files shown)


AlecAivazis:
One nice benefit of the interface is that it saves us the need to keep updating `AskOne` to accommodate the all of the different things one could do to a field.

core/write_test.go
```diff
@@ -252,44 +252,11 @@ func TestFindFieldIndex_tagOverwriteFieldName(t *testing.T) {
 	}
 }
 
-type testSettable struct {
-	Value string
-}
-
-type testSettableStruct struct {
```
coryb: The refactor lost the test on using custom types as a struct field, so I would add something like this back in:
```
type testSettableStruct struct {
    Settable testFieldSettable `survey:"values"`
}
```
Then in the `TestWriteWithFieldSettable` func add something like:
```
testSetStruct := testSettableStruct{}		
err = WriteAnswer(&testSetStruct, "values", "stringVal1")		
assert.Nil(t, err)		
assert.Equal(t, testSetStruct.Settable.Value, "stringVal1")
```

coryb:
Yeah, I am okay with pruning out the extra interface, I will update my project when released.  

If you changed to use a `Transform` field on `Question` then you would need to change `AskOne` to require a 4th param to pass the transform func in.  So it would be a breaking change.   Interfaces are used all over the go core, and I think the right solution for this problem.

AlecAivazis:
Yea I tend to agree with you - I'm going to add the extra test, thanks for the feedback!

Issue #82, opened 2017-07-22
Provide more examples, for all question types


ahmetb:
I was trying to use MultiSelect and tried to look for code examples, mostly because it's unclear to me how I'm supposed to collect the answers back (what `interface{}`'s actual type should be).

I tried `[]string`, didn't work. So I had to dig into [`examples/longlist.go`](https://github.com/AlecAivazis/survey/blob/master/examples/longlist.go), where I saw a survey.Select example. Apparently it deserializes into a `struct` that has a field that has the same name as `question.Name`. This sounds fine. 

But I _have to guess_ whether that field should be `[]int` or `[]string` because it's not documented at https://godoc.org/github.com/AlecAivazis/survey#MultiSelect and I can't find any examples. It seems like there's a little bit of friction or docs/examples gap.

ahmetb:
Please ignore the "I tried []string, didn't work" part. It was a validation error on my part (opening a separate issue).

AlecAivazis:
I'm not sure what sort of documentation you have in mind - I tried to document what the expected input was in the readme in the example for each prompt. I can try to put more information in the comment so that godocs picks it up, does that work for you? If you have a better idea for documentation, i'd love to review a PR if you submitted one 😉

AlecAivazis:
I"m going to close this - if you feel like I should add more documentation, please open another ticket :)

Issue #81, opened 2017-07-22
Document where answers can be written to


ahmetb:
I've been collecting answers to survey.MultiSelect etc. prompts into a `[]string`. But from the source code it looks like it's possible to use structs/maps too? https://github.com/AlecAivazis/survey/blob/6ee49f5e0a0ffadd78f5cd8c0b9429d84439846e/core/write.go#L19

If that's the case, Ask/AskOne methods that visibly accept `t interface{}` should probably document this behavior what `t`'s type should/can be.

AlecAivazis:
`WriteAnswer` handles the copying of the returned value from the prompt into whatever pointer you pass into `Ask`/`AskOne`. `[]string` is the right return type for a `MultiSelect` prompt, currently it does not support writing to anything else but I will address that as part of #84

AlecAivazis:
I"m going to close this - if you feel like I should add more documentation, please open another ticket :)

PR #87, opened 2017-07-23
Added response type docs for prompts

AlecAivazis:
@ahmetb, I added response type documentation for the various prompts in order to address your concerns in #82 and #81. Please let me know if there is something better you would like to see.

### DIFF
--- a/confirm.go
+++ b/confirm.go
@@ -9,7 +9,7 @@ import (
 	"github.com/AlecAivazis/survey/terminal"
 )
 
-// Confirm is a regular text input that accept yes/no answers.
+// Confirm is a regular text input that accept yes/no answers. Return type is a bool.
 type Confirm struct {
 	core.Renderer
 	Message string
--- a/confirm.go
+++ b/confirm.go
@@ -9,7 +9,7 @@ import (
 	"github.com/AlecAivazis/survey/terminal"
 )
 
-// Confirm is a regular text input that accept yes/no answers. Return type is a bool.
+// Confirm is a regular text input that accept yes/no answers. Response type is a bool.
 type Confirm struct {
 	core.Renderer
 	Message string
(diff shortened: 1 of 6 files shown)


ahmetb:
This is great. Thanks. 

Consider using "go test" ExampleXxx format to provide examples easily viewable in the generated godoc.

confirm.go
```diff
@@ -104,8 +104,14 @@ func (c *Confirm) getBool(showHelp bool) (bool, error) {
...
 
-// Prompt prompts the user with a simple text field and expects a reply followed
-// by a carriage return.
+/*
+Prompt prompts the user with a simple text field and expects a reply followed
+by a carriage return.
+
+	likesPie := false
+	prompt := &survey.Confirm{ Message: "What is your name?" }
+	survey.AskOne(prompt, &name, nil)
```
coryb: This one looks like a bad copy/paste, correct the prompt to something like `Do you like Pie?`, and use the `likePies` variable.

coryb:
I was going to suggest getting ExampleXxx tests working and started playing with it myself, but it is tricky if we wanted to use the `// Output: ` to test the output because of the terminal control characters that are printed.  We could make some Examples that just test syntax like this:
```
func ExampleConfirm_Prompt() {
    likesPie := false
    prompt := &Confirm{Message: "Do you like pie?"}
    AskOne(prompt, &likesPie, nil)
}
```
With this the godocs will have a magic example that shows up.

confirm.go
AlecAivazis: Nice catch - fixing it now on a separate branch.

AlecAivazis:
@coryb Sorry for not waiting for the review, I was feeling bad that there were no useful docs on the godoc page and figured we could get your changes in after the fact.

I fixed the typo on another branch, feel free to push onto that if you end up getting the examples going, otherwise I can try to get something together tonight when I get back from dinner.

coryb:
n/p, I was busy most of the weekend.  I probably wont have time to do get to the examples really in the next week, so if you feel like making the changes go ahead :)

AlecAivazis:
So I spent some time thinking about this further and I'm not sure what I think about the documentation being under `Prompt`, since that's sort of an implementation detail. I couldn't figure out how to provide a type-wide example, so I think I am in favor of keeping it how it is now - thoughts?

PR #90, opened 2017-08-20
Add feature `Editor`.

Equim-chan:
* Add `Editor` struct, referencing [Inquirer.js#editor](https://github.com/SBoudrias/Inquirer.js#editor---type-editor).
* Add `error.go` for defining `InterruptErr`.

### DIFF
--- a/editor.go
+++ b/editor.go
@@ -138,6 +138,7 @@ func (e *Editor) Prompt() (interface{}, error) {
 	cmd.Stdin = os.Stdin
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
+	terminal.CursorShow()
 	if err := cmd.Run(); err != nil {
 		return "", err
 	}
--- a/editor.go
+++ b/editor.go
@@ -1,0 +1,158 @@
+package survey
+
+import (
+	"bytes"
+	"errors"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"runtime"
+
+	"github.com/AlecAivazis/survey/core"
+	"github.com/AlecAivazis/survey/terminal"
+)
+
+/*
+Editor launches an instance of the users preferred editor on a temporary file.
+The editor to use is determined by reading the $VISUAL or $EDITOR environment
+variables. If neither of those are present, notepad (on Windows) or vim
+(others) is used.
+The launch of the editor is triggered by the enter key. Response type is a
+string.
+
+	message := ""
+	prompt := &survey.Editor{ Message: "What is your commit message?" }
+	survey.AskOne(prompt, &message, nil)
+*/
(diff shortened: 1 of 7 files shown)


AlecAivazis:
Thanks for submitting this @Equim-chan. I'll take a look over the next few days and hopefully we can merge it relatively quickly

error.go
```diff
@@ -0,0 +1,9 @@
+package survey
+
+import (
+	"github.com/AlecAivazis/survey/terminal"
+)
+
+var (
+	InterruptErr = terminal.InterruptErr
```
AlecAivazis: if we are going to define a special error for interrupts (not sure if its necessary tbh) then i would prefer to keep it in a single place, rather than having it spread across both packages.

editor.go
```diff
@@ -0,0 +1,168 @@
+package survey
+
+import (
+	"bytes"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"runtime"
+
+	"github.com/AlecAivazis/survey/core"
```
AlecAivazis: All of the imports to `survey` need to come from `gopkg.in/AlecAivazis/survey.v1` so that there is only one copy being used (allowing for editing the icons and what not)

error.go
Equim-chan: I think it's necessary, since interrupt is commonly treated as a signal of "quit". If the user wants to terminate the program at a survey query, he might send an interrupt signal. The upstream application relied on survey is responsible to react to the interrupt signal properly, whether quit the program or jump out or a loop, etc. Thus InterruptErr should be recognizable from other kinds of errors. How do you think?

editor.go
Equim-chan: OK, though I don't think it's a good idea, as it breaks the meaning of version control. It breaks anyone who wants to `go get` the package at master branch. (It seems like a known bug in gopkg.in)

error.go
AlecAivazis: Yea, i think it makes sense to have a specific error for interrupt that people can check against. Can we just have one in `terminal` and forgo the double declaration?

Equim-chan: Okay. I may commit later today.

editor.go
AlecAivazis: Hm so i'm still getting used to the proper patterns when pushing packages that rely on `gopkg.in` I will do some research to figure out what the right thing to do is. However, i think hosting the package through `gopkg.in` and being able to control which version is being pulled seems like it's worth this slight complexity. 

Let's make the imports consistent now and if we decide they need to go back to the normal `github.com/...` then we can change them all together

Equim-chan:
Done.

AlecAivazis:
Awesome - thanks for making those changes!

PR #92, opened 2017-09-05
Feat/nested result casts

AlecAivazis:
This PR extends the type handling support of `survey` to handle nested casts - for example the following is now supported:

```golang
// the questions to ask
var qs = []*survey.Question{
    {
        Name: "choices",
        Prompt:   &survey.MultiSelect{
            Message: "How old are you?",
            Options: []string{"1", "2", "3"},
        },
    },
}

func main() {
    // the answers will be written to this struct
    answers := struct {
        Choices           []int                     
    }{}

    // perform the questions
    _ := survey.Ask(qs, &answers)

    // answers.Choices should contain integer casted versions of the selected choices
}
```

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -221,14 +221,15 @@ func copy(t reflect.Value, v reflect.Value) (err error) {
 			// if its a slice
 			case reflect.Slice:
 				// an object of the correct type
-				obj := reflect.Zero(t.Type().Elem())
-				fmt.Println("from", v.Index(i), v.Index(i).Type().Kind(), "to", obj, obj.Kind())
-				// write the appropriate value to the obj
-				copy(obj, v.Index(i))
-				fmt.Println(">>", obj)
+				obj := reflect.Indirect(reflect.New(t.Type().Elem()))
+
+				// write the appropriate value to the obj and catch any errors
+				if err := copy(obj, v.Index(i)); err != nil {
+					return err
+				}
 
 				// just append the value to the end
-				reflect.Append(t, obj)
+				t.Set(reflect.Append(t, obj))
 			// otherwise it could be an array
 			case reflect.Array:
 				// set the index to the appropriate value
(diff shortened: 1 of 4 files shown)


core/write_test.go
```diff
@@ -154,6 +137,32 @@ func TestWrite_returnsErrorIfInvalidMapType(t *testing.T) {
...
+	err := WriteAnswer(&target, "name", []string{"1", "2", "3"})
+
+	// make sure there was no error
+	assert.Nil(t, err, "WriteSlice to Int Slice")
+	// and we got what we wanted
+	assert.Equal(t, []int{1, 2, 3}, target)
+}
+
+func TestWrite_writesStringArrayToIntArray(t *testing.T) {
+	// make a slice of int to write to
```
coryb: minor nit, should be:
// make an arrray of int to write to

core/write.go
```diff
@@ -202,7 +212,34 @@ func copy(t reflect.Value, v reflect.Value) (err error) {
...
+				// an object of the correct type
+				obj := reflect.Indirect(reflect.New(t.Type().Elem()))
+
+				// write the appropriate value to the obj and catch any errors
+				if err := copy(obj, v.Index(i)); err != nil {
+					return err
+				}
+
+				// just append the value to the end
+				t.Set(reflect.Append(t, obj))
```
coryb: Not sure, I wonder if the entire `reflect.Slice` section can be simplified to basically match you do in the `reflect.Array` case.  Perhaps something like:
```
case reflect.Slice:
    if t.Cap() < v.Len() {
        t.SetCap(v.Len())
    }
    fallthrough
case reflect.Array:
    copy(t.Slice(i, i+1).Index(0), v.Index(i))
```

AlecAivazis:
@coryb that was the first thing I tried and had some difficulty getting this to work right any other way. Would definitely be interested in seeing an alternative that passes the tests if you have one.  Thanks for taking a look at this!

PR #99, opened 2017-10-08
Term: Fix intensity attr on Windows

Equim-chan:
https://github.com/AlecAivazis/survey/blob/ced9eac915e96d8c1f2233d4bfc21c51db497b4d/select.go#L46
`green+hb` is actually `\x1b[1;92m`, but `terminal/output_windows.go` did not handle the range.

Before:
image

After:
image


AlecAivazis:
Thanks for submitting this @Equim-chan. Unfortunately I haven't gotten a chance to test this on my windows machine but I'll make sure to give it a go over the weekend.

PR #100, opened 2017-10-13
Fixed map example

crazcalm:
At first, I was really confused as to why the example code was not working
and then read the error. Quick fix. Thanks for the useful error msg!


AlecAivazis:
doh! thanks for submitting the fix!

Issue #98, opened 2017-10-08
Is there a way to pass a different value on a Select prompt?


mattstratton:
I have a Select prompt, and would like to display one thing to the user, but return a different value than what is being displayed.

That is to say, it will show "George Bluth", but would like it to instead return the value of "george-bluth", which is something i could have in a map, let's say.

In this particular case, I can transform the value myself (since it's simple toLower and string replacement), but in other cases it might be more arbitrary.

Frodgy:
Yeah, it would be much better to pass an array as options to return an index of answer.

AlecAivazis:
Hey @mattstratton! Currently there isn't a way to do this. I've been trying to think of the right API to support this. One approach would be to have another interface similar to `settable` which is described [here](https://github.com/AlecAivazis/survey#custom-types) which could be passed to the select. Or we could change the option to a struct instead of string to store the two values. 

What do you think?

Also,@Frodgy  FWIW, `Select` doesn't pass an index (nor `MultiSelect` a slice) so that a user could use a `Select`/`MultiSelect` to collect intsor an int slice depending. If you strongly think this should be changed, let's open up a ticket! I'd love to discuss it some more.

Frodgy:
@mattstratton, didn't you just say, that it would be a good idea to use struct to store more values instead of one string? At least, I got a case, when you have to select a book from an array, so with your interesting "survey" thing it goes in a really fancy way, except the dirty moment, when I have to search the returned string from survey across the whole book array.

dpastoor:
Couldn't you just add a `Callback` to the Question struct, similar to the Validator, that would be invoked like middleware.

```go
[]*survey.Question{
    {
        Name:     "name",
        Prompt:   &survey.Input{Message: "What is your name?"},
        Validate: survey.Required,
        Callback: func (val string) string {
                              return strings.ToLower(val)
                         },
    },...
```

Frodgy:
@dpastoor, is it real to use to validators at the same time? Like Required and MinLength(n)?

dpastoor:
that is something for @AlecAivazis to answer. Mine was just a hypothetical example, in which I pulled the survey.Question from the readme and added the Callback (hypothetical) extension.

AlecAivazis:
[...]

Issue #4, opened 2016-10-30
Add Transform to Question


AlecAivazis:
This attribute is a function that mutates the choice value into a different internal value

AlecAivazis:
Reopening to give @kataras the hacktoberfest credit 👍

kataras:
ahah thanks @AlecAivazis

PR #104, opened 2017-10-20
Implement the `Transform` idea

kataras:
As requested at https://github.com/AlecAivazis/survey/issues/98 and proposed at https://github.com/AlecAivazis/survey/issues/4 :)

## Results

> shown immediately


kataras:
@AlecAivazis You're welcome, yes `TransformString` can be used for all these but with helpers like `Title` and `ToLower` we give the user a first view of what they can do and how they can do it, I think it's better to let them as they are and not introduce any other helpers for strings.

PR #109, opened 2017-10-27
Add support for other in select and multi select

walle:
Add Other member to Select and MultiSelect structs.
If set to true another choice will appear, after all supplied options,
named Other, if selected the user is prompted to supply their input.
Used to have a set of choices but allowing the user to enter their own
value for the prompt.

This commit only includes tests for the output, not for any interaction.

This PR includes a proof of concept for the feature for discussion. For being a complete PR tests needs to be added for the functionality.

* Is this a feature you would like to have in the library?
* I'm not super happy about how the output looks, do you have any suggestions on how to improve it?
* Any other comments?

## Output example

This is one select and one multi select with choices of colors, with other choice added.

```
$ go run main.go
? Choose a color:
  red
❯ blue
  green
  Other

$ go run main.go
? Choose a color: blue
? Choose any color:
  ◉  red
  ◯  blue
  ◉  green
❯ ◉  Other

$ go run main.go
? Choose a color: blue
? Choose any color:
  ◉  red
  ◯  blue
  ◉  green
❯ ◉  Other
? Other pink

$ go run main.go
? Choose a color: blue
? Choose any color:
? Choose any color: red, green, pink
```

### DIFF
--- a/multiselect.go
+++ b/multiselect.go
@@ -30,6 +30,7 @@ type MultiSelect struct {
 	selectedIndex int
 	checked       map[string]bool
 	showingHelp   bool
+	Other         bool
 }
 
 // data available to the templates when processing
@@ -61,18 +62,24 @@ var MultiSelectQuestionTemplate = `
 
 // OnChange is called on every keypress.
 func (m *MultiSelect) OnChange(line []rune, pos int, key rune) (newLine []rune, newPos int, ok bool) {
+	// Create the choices array with other support
+	choices := m.Options
+	if m.Other {
+		choices = append(choices, "Other")
+	}
+
 	if key == terminal.KeyArrowUp {
 		// if we are at the top of the list
 		if m.selectedIndex == 0 {
 			// go to the bottom
-			m.selectedIndex = len(m.Options) - 1
+			m.selectedIndex = len(choices) - 1
 		} else {
 			// decrement the selected index
 			m.selectedIndex--
 		}
 	} else if key == terminal.KeyArrowDown {
 		// if we are at the bottom of the list
-		if m.selectedIndex == len(m.Options)-1 {
+		if m.selectedIndex == len(choices)-1 {
 			// start at the top
 			m.selectedIndex = 0
 		} else {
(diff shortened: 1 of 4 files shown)


AlecAivazis:
Thanks for the contribution @walle. I especially appreciate you starting the conversation with an actual PR of the feature - makes the discussion much easier 👍.

Since this will be a new addition to the top-level API I want to make sure we weigh the pros and cons before merging and taking the time to add the missing tests. My current understanding of this feature is that it adds an option that has a different color which the user can select through the usual Prompt's manner. 

I guess the first question is if the change in coloring is worth the additional API and I'm not sure I have a strong feeling either way. I will try to come to a conclusion soon but if you have a particular usecase that is served well by the differently colored option, I'd love to hear it.

Also would love @coryb and @kataras's opinion on this.

coryb:
It is interesting, but I am not sure it is really necessary.  If I had a need for the "other" functionality I think I would handle it with sequential prompts, something like this:
```
    var color string
    survey.AskOne(
        &survey.Select{
            Message: "Choose a color:",
            Options: []string{"red", "blue", "green", "other"},
        },
        &color,
        nil,
    )
    if color == "other" {
        survey.AskOne(
            &survey.Input{
                Message: "Enter other color:",
            }
            &color,
            nil,
        )
    }
    fmt.Printf("Color: %s\n", color)
```
```
$ go run ./color.go
? Choose a color:
  red
  blue
  green
❯ other
<enter>
? Choose a color: other
? Enter other color: pink
Color: pink
```

Would this approach work for you?

walle:
Thanks for the input guys! What this would do is basically embedding @coryb 's solution in the library. But for now I think the solution @coryb suggested works. Not sure the top level API needs to be modified. I'll close this now :)

kataras:
The PR is good and @walle seems to think hard before pushing this, but the balance of the feature it adds (? we can ask for Other without this feature as well) and its complexity doesn't feels right to me  (to check the m.Other everywhere is not an easy task, especially for any future additions).

kataras:
The PR is good and @walle seems to think hard before pushing this, but the balance of the feature it adds (? we can ask for Other without this feature as well) and its complexity doesn't feel right to me  (to check the m.Other everywhere is not an easy task, especially for any future additions), just a persomal note. People who need this they can just make a simple function in their  apps which will wrap the `survey.Ask/AskOne` and support the "Other" as well.

PR #110, opened 2017-10-29
allow to change the editor behavior

suzuki-shunsuke:
Add two fields `HideDefault` and `AppendDefaul` to `Editor`.

If `HideDefault` is `true`, the default value is not shown
before the launch of the editor. By default this value is `false`.

If `AppendDefault` is `true`, the default value is written to the temporary
file before the launch of the editor and `Prompt()` returns the empty
string if the contents of the temporary file is empty.
By default this value is `false`.

### DIFF
--- a/editor.go
+++ b/editor.go
@@ -26,9 +26,11 @@ Response type is a string.
 */
 type Editor struct {
 	core.Renderer
-	Message string
-	Default string
-	Help    string
+	Message       string
+	Default       string
+	Help          string
+	HideDefault   bool
+	AppendDefault bool
 }
 
 // data available to the templates when processing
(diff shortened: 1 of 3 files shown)


suzuki-shunsuke:
This change refers to [Inquirer.js](https://github.com/SBoudrias/Inquirer.js)'s editor behavior.

In many cases the default value is too long to show to the terminal.
And I hope the default value is written in the temporary file when the text editor launches.

AlecAivazis:
Thanks for the contribution @suzuki-shunsuke! Sorry for not getting back to you sooner. Would you mind adding tests to `tests/editor.go` aswell with examples of using both of those parameters?

suzuki-shunsuke:
Hi @AlecAivazis.

Thank you for reply!
I have added tests to tests/editor.go, but to be honest I don't know the good prompt messages.
If the prompt messages are not good, please tell me to correct them.

Best regards.

AlecAivazis:
@suzuki-shunsuke - Those prompts are fine - they're mostly used in automated tests so their values aren't critical.

Just verified all the behavior one last time. Will merge once travis gives us the 👍.

Thanks again for the contribution!

PR #123, opened 2018-02-26
boolean values pass through the required validator

### LINKED ISSUE #120: survey.Required validation not compatible with survey.Confirm prompt
Replicated behavior on both Windows & Linux using the following code:

```
package main

import (
  "fmt"
  "gopkg.in/AlecAivazis/survey.v1"
)

func main() {
  continue := false
  survey.AskOne(
    &survey.Confirm{
      Message: "Continue?",
    },
    &continue,
    survey.Required,
  )
  fmt.Println("continue", continue)
}
```

Expected to see output corresponding to the users input (true/false)

Actual behavior is if the user inputs n/N/No, an error is displayed: *"Sorry, your reply was invalid: Value is required"*.

AlecAivazis:
This PR fixes #120 by adding booleans as a special case in the `Required` validator. The logic behind this changed is describe in that issue, copied below
>  It's definitely an unexpected consequence of the zero value for a boolean being false. I think it would make sense to check if the type we are validating against is a boolean, and having it pass through survey.Required without causing an error. While this is slightly off of the documented behavior (since false is the zero value), I think it makes sense from the mindset that the prompt type might change but you requiring a response does not, and survey.Required should check that any answer was given, not just one of the zero value.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -8,8 +8,11 @@ import (
 
 // Required does not allow an empty value
 func Required(val interface{}) error {
+	// the reflect value of the result
+	value := reflect.ValueOf(val)
+
 	// if the value passed in is the zero value of the appropriate type
-	if isZero(reflect.ValueOf(val)) {
+	if isZero(value) && value.Kind() != reflect.Bool {
 		return errors.New("Value is required")
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Yea I wanted to avoid as many special cases as possible and keep `survey.Required` "simple" but i think there is a decent argument to be made for having it pass through.

Issue #121, opened 2018-02-21
[Feature Request] vim like j/k keybind at Select and Multi-select


tsuyoshiwada:
Hi @AlecAivazis, Thank you very much for the awesome library!

## Feature Request

I think that it is great to use if you can move up and down using `j/k` like vim keybind :clap:
Thanks!

## My Environment

```
$ sw_vers
ProductName:    Mac OS X
ProductVersion: 10.13.2
BuildVersion:   17C205
```

AlecAivazis:
Hey @tsuyoshiwada - thanks for the kind words.

I'm definitely interested in adding this. Starting to add support for vim keybindings sounds like a rabbit hole. If you think adding `j/k` to select would be enough, that's very doable.

tsuyoshiwada:
@AlecAivazis Thanks for the quick response!

> If you think adding j/k to select would be enough, that's very doable.

Yes! I think that moving up and down using by `j/k` is enough :smiley:
(Of course, hoping that the arrows so far will also work.)

AlecAivazis:
@tsuyoshiwada - I added support for navigating `Select` and `MultiSelect` with the `j` and `k` keys. Would you mind trying that PR out and running through both test suites to make sure it works?

tsuyoshiwada:
@AlecAivazis Thank you very much! Even if it is late, I will check the behavior today :)

tsuyoshiwada:
Hi @AlecAivazis, I confirmed the behavior of PR. It worked as perfect :tada:

AlecAivazis:
@tsuyoshiwada - thanks for checking! 

I just cut a new release that includes another small update I recently made. You should be able to get this release now with:

```
go get -u gopkg.in/AlecAivazis/v1
```

tsuyoshiwada:
Thx :smile: :tada:

PR #124, opened 2018-02-26
Added vim-like keybindings for navigating Select and MultiSelect

AlecAivazis:
This PR addresses #121  makes it so that j and k can be used to navigate select and multi select.


zanven42:
could we also allow h to deselect and l to select when in vim mode

PR #125, opened 2018-03-06
add filtering on multi-selects with lots of options

### LINKED ISSUE #61: usability issues for Select and MultiSelect with many options

jstrachan:
this fix allows a filter term to be typed which filters the list of options displayed.
users can use the delete/backspace key to delete characters from the filter and use Ctrl+W / Ctrl+X to clear the filter

fixes #61


jstrachan:
it looks like `k` and `j` can be used now for up/down but that would conflict with the filtering logic; I wonder if there needs some key press to switch filter v navigate modes? don't know vi well enough to recommend an approach though ;)

AlecAivazis:
@jstrachan - thanks so much for adding this. Like I said in the issue, I am at work right now and won't be able to review this until later tonight. 

That being said, I scrolled through the changes on my phone and nothing really stood out as problematic. That being said, I noticed you only added `Filter` to `MultiSelect`, can you add the logic to `Select` aswell? I'll take a deeper dive tonight and let you know if anything else pops up.

concerning the j/k select conflict: We're really starting to find ourselves down the "reproduce vim keybindings" rabbit hole that I wanted to avoid, but i think adding `esc` as a way to toggle between modes would be reasonable. Not sure how to best indicate the mode the user is currently in, but i'm open to suggestions

jstrachan:
Sure, I could add it to `Select` too - I figured I'd see if you're happy with the changes first.

I did wonder about using `esc` to clear the filter - but then found that seemed to go back in a survey I think; so not sure on the vim mode stuff. Maybe its a configuration option whether to be in vim mode or filter mode?

AlecAivazis:
So currently, if I send two `esc` in `Slect` i get a rather ugly error: `AskOne on standard's prompt failed: Unexpected Escape Sequence: ['\x1b' '\x1b'].no options`. If we could prevent that from happening by having special handling for the esc key, we could have 2 wins with a single feature 🎉 🎉 🎉 

I also checked `Select` and didn't see any behavior for the `esc` key, not sure what you mean by "that seemed to go back in a survey"

AlecAivazis:
Just to add some more info around my thinking: having j/k be opt in is a slightly weird user experience for the end user. Seems less than ideal for the developer to decide if they want to support vim-like keybindings or not since it's so dependent on their users preference.

I think ideally, we would always support filtering, since it does speed up the user's ability to find their option, even for short lists. 

Thoughts?

jstrachan:
@AlecAivazis I guess the user of Survey could enable/disable vim mode in the Survey? Then we could have a key press to switch into and out of vim mode? Certainly from my own use case I'm not really too interested in having a vim mode - am happy with the cursor keys & really need filtering :)

AlecAivazis:
Yea i think providing a way for the end user to toggle between vim-mode and filtering is what I"m in favor of. Having `esc` be the key that does that toggling sounds good to me from an "emulate vim" standpoint. That would allow us to always have filtering on, and a way to keep the vim keybindings working.

jstrachan:
OK the latest commit defaults to filter mode but allows the user to clear the filter with `Esc` or to toggle into/out of `vim mode` via the `Esc` key - for both `select` and `multiselect`

Though I have found that you have to hit `Esc` twice for it to register the `Esc` key. 

It turns out to be this code causing it - its parsing `Esc` and then `A, B, C, D` for the arrow keys
https://github.com/jenkins-x/survey/blob/4657d803ce6e9d9a549e972d991aa65e5f38818c/terminal/runereader_posix.go#L59

Given `Esc` is used to handle the `Esc, [, A/B` runes for up/down am not sure we can process a single `Esc` character.

I tried patching the code to this:
```go
func (rr *RuneReader) ReadRune() (rune, int, error) {
	r, size, err := rr.state.buf.ReadRune()
	if err != nil {
		return r, size, err
	}
	// parse ^[ sequences to look for arrow keys
	if rr.lastRuneEscape {
		if r != '[' {
			return r, size, fmt.Errorf("Unexpected Escape Sequence: %q", []rune{'\033', r})
		}
		r, size, err = rr.state.buf.ReadRune()
		if err != nil {
			return r, size, err
		}
		rr.lastRuneEscape = false
		switch r {
		case 'D':
			return KeyArrowLeft, 1, nil
		case 'C':
			return KeyArrowRight, 1, nil
		case 'A':
			return KeyArrowUp, 1, nil
		case 'B':
			return KeyArrowDown, 1, nil
		}
		return r, size, fmt.Errorf("Unknown Escape Sequence: %q", []rune{'\033', '[', r})
	}
	rr.lastRuneEscape = r == '\033'
	return r, size, err
}
```
which works in that a single `Esc` key press can now clear a filter or toggle between vim mode and non vim mode - the downside is when you have a filter, you can't use the up/down arrow keys - as that sends an Esc key and then clears the filter.

So for now we need to hit Esc twice; unless we can figure out how to get the ReadRune() function on posix to return a single rune for up/down cursor keys  and the `Esc` key?

jstrachan:
Aha - by just checking if there are any more bytes buffered we can detect if its a `Esc` key or an arrow key! OK so the double `Esc` issue is now fixed - yay!

AlecAivazis:
sweet!! Alright finally pulling this down and making sure things work. I also want to give it a try on my windows machine so i won't be able to finalize the review until tomorrow.

multiselect.go
```diff
@@ -27,6 +27,9 @@ type MultiSelect struct {
 	Default       []string
 	Help          string
 	PageSize      int
+	VimMode       bool
```
AlecAivazis: This field isn't needed anymore, right?

multiselect.go
```diff
@@ -27,6 +27,9 @@ type MultiSelect struct {
...
+	FilterMessage string
```
AlecAivazis: I'm curious what the FilterMessage is in the cli you are working on.

I'd like to provide some instructions to the user, indicating what mode their in. My original thinking was that when we are filtering, we can show the filter message along with the current filter value. A default message might be something like `You can type to filter items`. And when the user is in vim-mode, we could provide a message like `Use the arrow keys or j/k to highlight an option`.

multiselect.go
```diff
@@ -81,20 +87,51 @@ func (m *MultiSelect) OnChange(line []rune, pos int, key rune) (newLine []rune,
...
+			} else {
+				// otherwise just invert the current value
+				m.checked[options[m.selectedIndex]] = !old
+			}
 		}
 		// only show the help message if we have one to show
 	} else if key == core.HelpInputRune && m.Help != "" {
 		m.showingHelp = true
+	} else if key == terminal.KeyEscape {
+		if m.filter != "" {
```
AlecAivazis: Since we already provide a way to clear the filter, I think it would be best to not overloading the esc behavior. If the user  presses escape, i think we should keep the filter as is, and allow them to navigate the select with j/k. Is that too complicated of an interaction? Knowing that `esc` toggles the input, it was sort of confusing to see the filter disappear when I pressed it.

jstrachan:
@AlecAivazis no worries! :) I've added a couple of example apps that make it easy to test out filtering on long select / multi-selects:

* https://github.com/jenkins-x/survey/blob/b56313429abebbe8861dafce68cb7b34fe8dbdf5/examples/countrylist.go
* https://github.com/jenkins-x/survey/blob/b56313429abebbe8861dafce68cb7b34fe8dbdf5/examples/longmulti.go

multiselect.go
AlecAivazis: ^ Does mesh well with my other comment of leaving the filter behind when the user presses `esc`. I worry that we will run into overflow issues (which are only really a problem in windows per #101) but I think if we can find a concise way to provide good instructions, we'd be in awesome shape

multiselect.go
jstrachan: Yeah - the VimMode keeps track of if you're in vim mode or not (so its toggled by the `Esc` key). It also lets users start in Vim mode if they wanted

multiselect.go
jstrachan: so `FilterMessage` currently is `" " + m.filter` - i.e. its the thing shown on the screen after the input message.  I guess the template could just conditionally output `" " + m.Filter if m.Filter != ""`

jstrachan: BTW we clear `m.filter` after the user accepts the input so it is removed from the screen

multiselect.go
jstrachan: I felt that `Esc` was the most natural way to clear the filter; I wasn't sure how many users would think of hitting `Ctrl+W` or `Ctrl+X`

multiselect.go
AlecAivazis: Hmm allowing the developer to specify starting in vim-mode is an interesting idea. I wonder if that is something a developer should be able to decide for their users or if we should just default to being not in vim-mode.

jstrachan: I don't have a strong feeling either way TBH - only for my use cases I want to be in 'filter' mode OOTB :) But am sure a vim fan might disagree!

multiselect.go
AlecAivazis: Yea i understand that point, I think since survey already supports j/k for movement, like many other similar packages, I'm trying to find the right balance. While I agree that `ctrl+x` and `ctrl+w` are not super obvious, I want to keep the mental model for the original feature-set as clear as possible. I think having `esc` not clear the filter is the best way to do that. Like I said, as a vim user myself, I was very confused when esc cleared the filter.

multiselect.go
AlecAivazis: I think having vim mode be off by default is fine and we can make `VimMode` a private field. As part of another PR, i'd like to improve the messaging in the select and MultiSelect prompts to do 2 things:
1. Indicate to the user when there are options above and below the current cursor
1. Notify them of which mode they're in and general instructions for how to navigate.

I'm curious what @tsuyoshiwada thinks about this discussion.

multiselect.go
AlecAivazis: Ah so you're not expecting it to be modified by the user? Should it be a private field then?

jstrachan: I figure d a user may want to configure the default - but I'm not too worried if you want to make it private

multiselect.go
jstrachan: Yeah - I did wonder with the new magic `ctrl+x` and `ctrl+w` key presses we may need 'keys' help that appears (in addition to any custom help) to help users learn the key presses available?

multiselect.go
AlecAivazis: I suppose there’s no harm in extra configuration. Let’s leave it public. Sorry to be so nit picky. I love the feature and want to make sure we think through all of its impact on both developers and end users.

jstrachan: No worries! TBH I'd not even considered vim mode before this PR started so its definitely well worth trying to figure out how all these pieces fit together to get the best UX

multiselect.go
jstrachan: Or to say that another way. If I'm using some CLI tool thats using Survey under the covers and I start filtering a long list; how can I learn the magic `ctrl+x` /  `ctrl+w` key presses to clear the filter? Or learn that there's a vim mode. Figure we should have some way to show some help for the supported keys.

Due to screen real estate it might be we have to link to a little doc via a URL if we can't keep it pretty concise (a couple of lines of the terminal)?

AlecAivazis: Yea this is the exact concern I alluded to in the thread around the VimMode field. We already support Help text with every prompt. Wouldn’t be technically challenging to add navigations docs to those templates in the select/multi select prompts but finding the right copy might be trickier. 

I think whether esc clears the filter or toggles vim mode, we still have this documentation problem. How do you feel about having esc behave like it would in vim, and we address the docs propel either in this PR or another

jstrachan: sounds good to me!

jstrachan: BTW letting users know they can filter the list might be handy in the help too - it might not be obvious to new users

AlecAivazis: Yea I agree, this need is partly captured in #61, probably worth adding some more info. Maybe even a new ticket that isn't cluttered by all of the other usability improvements that have been made since it's opening.

jstrachan:
@AlecAivazis anything else you need me to do with this PR?

AlecAivazis:
@jstrachan - did u change the esc behavior to toggle between vim mode and leave the filter as it’s current value?

jstrachan:
@AlecAivazis ah sorry - forgot that ;) Just pushed a fix so that `Esc` only toggles vim mode and doesn't affect the filter

AlecAivazis:
Awesome, just confirmed the change. I'll merge this in right away.

jstrachan:
@AlecAivazis many thanks!

AlecAivazis:
@jstrachan thank *you* for implementing this. I should have a release cut by the end of the day so you should be able to grab it locally very soon.

jstrachan:
@AlecAivazis really looking forward to the new release, thanks! :)

AlecAivazis:
@jstrachan - sorry for not publishing that release yesterday. I wanted to add a little instructions to guide the user but couldn't quite find something that i liked. I'm going to cut the release now and we can add the instructions in a later patch.

AlecAivazis:
`v1.5.0` is now released. Thanks again @jstrachan!

Issue #133, opened 2018-03-22
[feature] Allow separate display and value fields for Select & MultiSelect


djgilcrease:
I was thinking something like
```go
type Option struct {
	Display string
	Value interface{}
}
```
this will also allow the value to be a struct not just a string

AlecAivazis:
They @djgilcrease - thanks for submitthing this issue. This seems to be a duplicate of #105 which has a lot of discussion about how one should specify an option as a struct.

PR #134, opened 2018-03-23
Discussion: Add Selection interface to support display value and a selection value

djgilcrease:
This is an API change that changes Select and MultiSelect to allow separation of display value and potential complex value of the selection. The primary usecase I see for this is pulling options from a REST api and having the Option.Value be the struct you parse out of the API while the Display being enough details for a user to know what they are selecting (see examples/complex.go)

Addresses: #133


AlecAivazis:
Hey @djgilcrease - thanks for submitting this PR. I am a bit uneasy changing the API this drastically since it will definitely be a new major version - can you provide some examples of how the current API doesn't support what you're looking to do?

djgilcrease:
Take a look at the examples/complex.go file, that is what I am trying to
do.

AlecAivazis:
Yea, i see the example you added - thanks for doing that.

I don't see how the  this new API is required in order to achieve this, however. It seems like going a much simpler route like updating options to take an interface would achieve what you want and not break the API. Am I missing something?

djgilcrease:
The full api change is likely not need, I will take a look at it again with
just changing the options field.

PR #135, opened 2018-03-23
Fixed special keys and terminal overflow

### LINKED ISSUE #131: keys HOME, END, DEL don't work as expected on UNIX
In [rundereader_posix.go](https://github.com/AlecAivazis/survey/blob/master/terminal/runereader_posix.go) 
there is no mapping for the keys HOME, END, DEL

If you press one of the symbols (Home, End, Del or any other of the command keys  e.g. Page Up, Page Down) 
`return r, size, fmt.Errorf("Unknown Escape Sequence: %q", []rune{'\033', '[', r})` is triggered as there is no mapping.

stefanvassilev:
When handling single line inputs the library doesn't respect delete, end, or other special keys. It just interrupts the process and exits. 

This adds Home, End, Forward delete functionality as well as improved handling of longer single line inputs ( stretching more than one line in the terminal window )

fixes #131

### DIFF
--- a/terminal/cursor.go
+++ b/terminal/cursor.go
@@ -61,16 +61,16 @@ func CursorHide() {
 func CursorMove(x int, y int) {
 	fmt.Printf("\x1b[%d;%df", x, y)
 }
+
 // CursorSave saves the current position
 func CursorSave() {
 	fmt.Print("\x1b7")
 }
 
-// CursorSave restores the current position of the cursor
+// CursorRestore restores the saved position of the cursor
 func CursorRestore() {
 	fmt.Print("\x1b8")
 }
-
 
 // CursorLocation returns the current location of the cursor in the terminal
 func CursorLocation() (*Coord, error) {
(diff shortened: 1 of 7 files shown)


AlecAivazis:
Thanks for submitting this @stefanvassilev - I'll need to test this on my windows machine to verify that everything works so it might take me a bit. Since you are dealing with terminal overflows, do you have any ideas how we could fix #101 on windows machines?

AlecAivazis:
Hey @stefanvassilev - I tried running the input tests and the cursor seems to start on the line above where it should:
screen shot 2018-03-26 at 9 10 23 am

Also, when I typed a very long answer, it seems to also have issues on cleanup:
screen shot 2018-03-26 at 9 10 52 am

AlecAivazis:
Hey @stefanvassilev  - i just tried the input tests again and the cursor starts on the right line but doesn't seem to clean up properly. Here's a video showing the behavior on a very thin terminal. It's slightly confusing since the `Answered: <foo>` response in the tests make it look like the answer duplicated a lot more, but it still looks like there is some duplication on the same line as the blue text where the template shows the answer: 

[Untitled.zip](https://github.com/AlecAivazis/survey/files/1860649/Untitled.zip)

Issue #85, opened 2017-07-23
Allow users to set output stream


ahmetb:
It users like this library is printing interactive dialogues to stdout. This is making it difficult to write programs that prompt questions, which then prints something to stdout users can redirect to a file and use.

e.g. A program like

    keyutil generate --to-file=- | jq .

Would print to stdout, which then would be consumed by the next program in the pipeline. However currently all questions go to stdout, making this impossible.

karuppiah7890:
What should be the output (to file or stdout) ? The questions and the answers together ? Or just the answers ? And in case of questions - the complete message of the question or just the name of the question which is used to identify the question and get the answer currently

ahmetb:
so answers are already in STDIN, they don't need to be in any of the output streams.

however, the questions, animations etc, they can go to STDERR (hopefully we're not expecting people to save ansi escape sequences we use to do animations to files).

this way I can easily write a program that only prints to STDOUT from my code and not from the `survey` package.

ahmetb:
I still need this feature and may take a look at it. Or would you like to tackle it yourself @AlecAivazis ? It looks like all the changes required would be in this repo.

AlecAivazis:
hey @ahmetb! Sorry for not responding - if you want to submit a PR with some test cases I would love to review it. I agree that sending only the results to `stdout` is probably the way to go. sending the animation stuff to stderr feels a bit weird but I'm not sure what else to do...

Looking forward to seeing it implemented!

AlecAivazis:
I'm going to close this since #143 has been merged. If it doesn't address your exact needs @ahmetb, feel free to open another issue

PR #139, opened 2018-05-03
Extract os.Std* as variables for testability

izumin5210:
Thanks for creating this great package!

Currently,  this package uses `os.Stdin` and `os.Stdout` directly in functions. So we cannot replace them with mock implementations.

### DIFF
--- a/confirm.go
+++ b/confirm.go
@@ -49,7 +49,7 @@ func yesNo(t bool) string {
 }
 
 func (c *Confirm) getBool(showHelp bool) (bool, error) {
-	rr := terminal.NewStdRuneReader()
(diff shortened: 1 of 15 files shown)


AlecAivazis:
Hey @izumin5210! thanks for submitting this. I think this is an important first step towards being able to write unit tests for surveys in a reasonable manner. There was some discussion in #85 concerning how to handle the escape sequences used for animation. I'm curious if you have any ideas. I'd like to see if there's a good answer to the issue discussed which we could merge as part of this.

AlecAivazis:
I'm going to close this to keep the repo tidy, feel free to reopen it when you have a chance to make the requested changes

PR #143, opened 2018-06-12
Refactor survey to use optionally specified stdio.

hinshun:
We wanted to refactor out survey's stdin/stdout to optionally use a different pair of files. The refactor was doing using the [functional options pattern](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis). The main purpose was to automate testing of interactive applications at Netflix, and one of which used this library. In order to demonstrate why having controllable stdio is useful, I've also included a full set of integration tests using https://github.com/Netflix/go-expect that leverages this refactor.

FAQ
---
- What is `go-expect`?
  - It's a library that can expect sequences from a output stream and send data into an input stream byte for byte, so that one can automate testing of an interactive application.
- What is `vt10x`?
  - It's an in-memory terminal emulator fulfilling the VT10X specifications, so that it can respond to e.g. cursor position report of its in-memory cursor. `go-expect` can multiplex the escape sequence from `survey` to `vt10x` and multiplex the cursor position report back to `survey`.
- Why modify `CursorLocation`?
  - Bytes sent in while the prompt is rendering and before the cursor position report is discarded due to the way that it was implemented. Unmatched bytes are saved to a `BufferedReader` which is used in `RuneReader` so that it will first read from this internal buffer before it's stdin. Before this change, if you send the character `"R"` before the cursor position report, it will hang indefinitely.
- Why vendor?
  - If a golang project imports any packages that has a vendor directory, go doesn't allow using nested dependencies from your `GOPATH`. It's also important for the library to be pinned to a set of versions (that consumers not necessarily have to match exactly) to have reproducible tests and for vendoring programs (like `dep`) to solve dependency constraints.
- Why not autoexpect or some other expect library?
  - Autoexpect isn't reliable, and `Netflix/go-expect` doesn't spawn processes to control them. It only interacts with the pty that it creates. This means that `go-expect` tests can be regular go `_test.go` files and can leverage golang's test features (race, coverage, parallelism).


coryb:
Not sure this is better ... but there is a way we can leave the Prompt interface the same by isolating the stdio into the core.Renderer object.  It does require a magic interface though.

So we could change `type Renderer` to have a `stdio terminal.Stdio` data member.  Then we could add a function to set it:
```
func (r *Renderer) WithStdio(stdio terminal.Stdio) {
    r.stdio = stdio
}
```

This would allow us to remove `Stdio` from `Error` and  `Render` function signatures (instead access via `r.stdio`).  Then the only calls that I see which require a `Stdio` is `terminal.NewRuneReader()` so I think we can add a helper function in Renderer to pass the stdio like:

```
func (r *Renderer) RuneReader() *terminal.RuneReader {
  return terminal.NewRuneReader(r.stdio)
}
```

So in confirm where we have:
```
    rr := terminal.NewRuneReader(stdio)
    rr.SetTermMode()
    defer rr.RestoreTermMode()
```
it would simple migrate to:
```
    rr := c.RuneReader()
    rr.SetTermMode()
    defer rr.RestoreTermMode()
```

The less-than-ideal part is that we would have to call the `WithStdio` inside the `Ask` code, but this can be done by checking if the prompt fits an interface that allows WithStdio:
```
type wantsStdio interface {
  WithStdio(terminal.Stdio)
}

// later in Ask
if t, ok := q.Prompt.(wantsStdio); ok {
  t.WithStdio(options.Stdio)
}
```

confirm_test.go
```diff
@@ -55,17 +59,103 @@ func TestConfirmRender(t *testing.T) {
 		},
 	}
 
-	outputBuffer := bytes.NewBufferString("")
-	terminal.Stdout = outputBuffer
-
 	for _, test := range tests {
-		outputBuffer.Reset()
+		r, w, err := os.Pipe()
+		assert.Nil(t, err, test.title)
```
coryb: nitpick, `assert.Nil(t, err)` should be written as `assert.NoError(t, err)`

editor_test.go
```diff
@@ -73,17 +79,148 @@ func TestEditorRender(t *testing.T) {
...
+			Editor{
+				Message: "Edit git commit message",
+			},
+			func(c *expect.Console) {
+				c.Expect("Edit git commit message [Enter to launch editor]")
+				c.SendLine("")
+				go c.ExpectEOF()
+				time.Sleep(time.Millisecond)
+				c.Send("iAdd editor prompt tests\x1b")
+				c.SendLine(":wq!")
```
coryb: The editor tests are failing for me since I have my EDITOR set to emacs.

coryb: I get a little further with by forcing the EDITOR, but the DSR codes seems to be going to my console, and the tests hang for me after that:
```
$ EDITOR=/usr/bin/vi go test
^[[47;178R^[[31;22R^[[31;23R^[[31;24R^[[31;25R^[[31;26R^[[31;27R^[[31;28R^[[31;29R^[[31;30R^[[31;31R^[[31;32R
<hangs>
```

Not sure if it is a OSX thing or if there is a race condition with the tests.  When I look at the process table, there is no vim subprocess, so it looks like it getting past the editor tests.

coryb: I narrowed the hanging test to: `TestMultiSelectPrompt/Test_MultiSelect_prompt_interaction_with_vim_mode`

So I suspect vim-mode require some external configuration?  So might not be related to vt10x testing.

hinshun: There was an issue with runes not being fully read via `ExpectEOF`. This was fixed in: https://github.com/Netflix/go-expect/commit/517479316a06a1ce93e1e22bc7f2e7a8f9713dc2

terminal/runereader.go
```diff
@@ -34,43 +34,48 @@ func (rr *RuneReader) ReadLine(mask rune) ([]rune, error) {
...
+	terminalSize, _ := cursor.Size(rr.state.buf)
 	for {
 		// wait for some input
 		r, _, err := rr.ReadRune()
 		if err != nil {
 			return line, err
 		}
 		// we set the current location of the cursor and update it after every key press
-		cursorCurrent, err := CursorLocation()
+		cursorCurrent, err := cursor.CursorLocation(rr.state.buf)
```
AlecAivazis: Small nitpick. Since the cursor functions are now methods of a struct, can we remove the `Cursor` at the front of all of these? So this would look like `cursor.Location(rr.state.buf)`?

AlecAivazis:
I would much prefer it if we could leave the interface for `Prompt` the same as before, per @coryb's suggestion. 

Although I do not know of anyone with custom Prompts, my gut says that a change to a core interface like `Prompt` is a public-API breaking change and would deserve bumping the major version of `survey`, something I"ve been trying to avoid for a while. 

I think it's perfectly reasonable to modify the internals of `Ask` to accommodate this ability to specify stdout/in.

coryb:
We are planning on testing Windows today, went in a few circles yesterday testing OSX.  The OSX tests are working.  

There is an issue (not new to this PR) where the test don't work unless you set `EDITOR=vi` (it forks of emacs for me and hangs), we need to fix this (later PR) to remove the `init()` function that overrides the `editor` variable so it can be modified for testing.

The second odd issues is that the tests are echoing the raw CSR codes to the terminal which is why I see all the DSR codes `^[[47;178R^` echo'd to the terminal when running the tests.  I think we will work on escaping the test output (ie `%q`) so we can see the encoded output, not the raw csr output.

terminal/runereader.go
hinshun: I agree. I've also been thinking about an alternative where `CursorUp` and the like simply return `string` so that the caller can write to appropriate writer, but `CursorLocation` will still need to interact with the tty directly. If we stay with the `struct` approach I definitely think we should remove the redundant `Cursor` in the methods.

hinshun:
I've made @coryb changes as suggested so that `Prompt`, `Cleanup`, `Error` remain unchanged interfaces.

Note that because windows does not support creating psuedoterminals, the newly added integration tests are skipped on windows. The full suite passes on both OSX & Linux, and I've manually went through Windows testing on `cmd.exe`. In order to get the test suite running on windows too, we'll have to investigate how to emulate a pty if its necessary, simply have `go-expect` return a file from `os.Pipe()`, or some solution involving [winpty](https://github.com/rprichard/winpty). I'm unsure what's the level of work necessary, but it's a low priority for me atm.

The EDITOR problem @coryb was fixed, I realized the `Editor` struct accepts a `Editor` field to override the binary to execute, so they were hardcoded for `vi` and there is a check that skips the tests if `vi` isn't found.

The CSR codes echoing to terminal problem is also fixed, they occurred because when writing the test output (which includes CSR codes) your emulator also responds by writing to stdin. Tests now output the raw `%q` escaped strings as well as a dump of runes in the virtual terminal's screen.

I think consensus has changed around library vendoring with the arrival of [dep](https://github.com/golang/dep) solving dependency constraints, but if you feel strongly about it I can remove it.

AlecAivazis:
I'm really excited you were able to keep the user-facing interfaces the same. Since you checked windows and say that works, it sounds like you've addressed most of my high-level requirements. I'll try to find some time tonight to dig into the actual changes and run the tests on my windows machine. 

I would love to be able to run the integration tests through a windows machine on travis to ensure those don't break. But without going through the actual changes, i'm not sure what the level of effort looks like there. 

I'll report back after i've run through the tests.

In the meantime, could you add a section to the readme showing users how to write their own tests?

hinshun:
I've added a section to the README about writing `go-expect` tests.

AlecAivazis:
I can't seem to find an explanation for what `dep` does differently that caused the shift in library vendoring conventions. Do you have a link you could share? Or maybe a short explanation? I'm not opposed to vendoring our dependencies if its the "right" thing but I want to make sure that we're not going to impact users negatively

multiselect_test.go
```diff
@@ -89,17 +92,149 @@ func TestMultiSelectRender(t *testing.T) {
...
+			"Test MultiSelect prompt interaction overriding default",
+			&MultiSelect{
+				Message: "What days do you prefer:",
+				Options: []string{"Sunday", "Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday"},
+				Default: []string{"Tuesday", "Thursday"},
+			},
+			func(c *expect.Console) {
+				c.ExpectString("What days do you prefer:  [Use arrows to move, type to filter]")
+				// Deselect Tuesday.
+				c.Send("\x1b[B\x1b[B ")
```
AlecAivazis: It should be possible to use the internal `survey/core` methods on a Cursor that points to the `expect.Console`'s stdio to abstract away the actual ansi escape sequences here right? It's not a huge deal but it would definitely make it friendlier for future developers to write these tests if they didn't have to look up these values.

coryb:
fwiw, I think the vendoring is okay.  Even in the older document you linked to it says:
> You can carve out an exception for yourself if your library has hermetically sealed its dependencies, so that none of them escape to the exported (public) API layer. No dependent types referenced in any exported functions, method signatures, structures

Since none of the vendored libraries are exposed in our API this should be fine.   Also when users use `dep` it will prune away library `vendor` directories (by default) anyway, so whether we vendor or not it is irrelevent to others using vendoring.

The maintainers of `dep` think it [should be used for libraries].(https://github.com/golang/dep/issues/208)

We can always re-evaluate if we determine it is a problem, but I dont forsee any problems currently.

hinshun:
I've removed `autoplay` and updated README and CONTRIBUTING accordingly.

AlecAivazis:
Ah great thanks @coryb. Totally on board with vendoring our deps.

multiselect_test.go
AlecAivazis: Ah darn well that's definitely good enough for now

AlecAivazis:
Hey @hinshun - is there anything else you want to add to this PR? I'd like to get it merged ASAP

AlecAivazis:
Great - i think this is good to go so i'm going to merge it. Thanks again for all the hard work @hinshun!

AlecAivazis:
This should be available now in `goppkg.in/AlecAivazis/survey.v1`

PR #149, opened 2018-07-30
Use shellquote to split EDITOR before executing the editor prompt

### LINKED ISSUE #147: Split content of EDITOR variable on whitespace
One of the users of an application that uses this library reported this issue:

```
Prompt [Enter to launch editor] Error: exec: "emacs -nw": executable file not found in $PATH
```

The issue is that their `EDITOR` variable is set to `emacs -nw` and `survey` currently uses the entire content of the variable as the executable name (https://github.com/AlecAivazis/survey/blob/master/editor.go#L156).

In shells like `bash`, line splitting is on by default so this "just works" in most cases.

This issue can be worked around by having people set their `EDITOR` variable to be a script that invokes the editor they want with the arguments that they want, but that can be a bit of a hassle. However, making this change in code here could be messy as well (determining exactly how to split on whitespace etc.).

Wanted to file this issue to see if others have encountered this or have feedback on it.

hinshun:
Closes #147

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -20,6 +20,12 @@
   revision = "1954e646417484a2a687ea344edade2c2b6523c8"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/kballard/go-shellquote"
+  packages = ["."]
+  revision = "95032a82bc518f77982ea72343cc1ade730072f0"
+
+[[projects]]
   name = "github.com/kr/pty"
   packages = ["."]
   revision = "282ce0e5322c82529687d609ee670fac7c7d917c"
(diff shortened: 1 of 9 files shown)


hinshun:
Needs testing on windows.

AlecAivazis:
Sorry for lagging on this @hinshun. I'm going to try to test it on my windows machine this weekend.

AlecAivazis:
I was able to verify that the test case passes locally on my laptop, I'm really not sure how to test this on windows so i'm going to merge this and we can always fix it later if someone reports it broken on windows.

Thanks again @hinshun!

Issue #93, opened 2017-09-07
Use more compatible symbols by default


Noah-Huppert:
The default [customization option values](https://github.com/AlecAivazis/survey#customizing-output),  while nice looking, are not extraordinarily compatible with the average user's font face. This is especially relevant for a library like survey. Which will probably be used in applications which interact with less technical users.  

It would be nice if survey shipped with some more compatible defaults:

```golang
core.ErrorIcon = "X"
core.HelpIcon = "????"
core.QuestionIcon = "?"
core.SelectFocusIcon = ">"
core.MarkedOptionIcon = "[x]"
core.UnmarkedOptionIcon = "[ ]"
```

They still end up looking pretty nice looking:

![2017-09-07-150747_1920x1080_scrot](https://user-images.githubusercontent.com/3323504/30180644-91416ba2-93de-11e7-93f4-61e604713655.png)

AlecAivazis:
That's a good point @Noah-Huppert. I definitely see the reason for `survey` to err on the side of compatibility here and use "icons" that are supported by any terminal. If people want something fancier, they are always able to change it themselves. If you want to submit a PR, I would gladly look it over, otherwise, I'll try to get something out this weekend.

AlecAivazis:
Since I'm on a windows machine, I figured I would leave a note that this also affects the standard `cmd.exe` terminal on windows and results in a pretty ugly experience.
<img width="274" alt="capture" src="https://user-images.githubusercontent.com/916317/31804459-5bc8387e-b50e-11e7-95a8-f6264875d4fd.PNG">

AlecAivazis:
Closing this since #150 addresses it

PR #150, opened 2018-08-10
Change icons to more simple ones (Issue 93)

MarkusFreitag:
While using your, by the way very convenient and easy to use, library in some CLI tools, I stumbled across the open issues.
So I've finished Noah-Huppert's work #93  and changed all icons to the proposed versions. I also have adjusted the tests to use string formatting instead of the hard-coded symbols. So in case we need to change the icons again in the future, it will be pretty easy.

### DIFF
--- a/confirm_test.go
+++ b/confirm_test.go
@@ -2,6 +2,7 @@ package survey
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"testing"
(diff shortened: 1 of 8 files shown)


AlecAivazis:
Hey @MarkusFreitag - thanks for putting this together!

I am currently incredibly busy with a client, so i'm not sure when I'll be able to get to this but I promise it will be as soon as I can. The changes look relatively innocuous, I just have to step through the testings to makes sure nothing weird has happened. Will definitely let you know if I have any more requests/updates.

Issue #126, opened 2018-03-07
Terminal cursor move/delete does not handle multi-character inputs well 


mritd:
When using the Input component to enter other language characters such as Chinese, the cursor does not move properly when characters are deleted.

![2018-03-07 21_08_38](https://user-images.githubusercontent.com/13043245/37093900-fc6dc848-224b-11e8-8d03-ab1bdae62d1a.gif)

Because [CursorBack function](https://github.com/AlecAivazis/survey/blob/6c118e265bb2c74e38a987746f763c34d5947769/terminal/runereader.go#L62) ignores the utf8 length of the last character, making the cursor always move one character to the left; for Chinese Character cursors should be moved left by 2 characters, but I do not know what other languages should do.

AlecAivazis:
Ah darn! Thanks for bringing this to my attention @mritd! Hopefully this can be addressed soon

mritd:
@AlecAivazis It seems to solve this problem through this library https://github.com/mattn/go-runewidth 😀

mritd:
@AlecAivazis 

I tried to fix this with the go-runewidth library and created a PR, now it looks fine.

refs #180

mritd:
Close this issue.
refs #217

PR #152, opened 2018-08-24
Length counter fixed for emoji like multi byte characters

ozguncagri:
In golang len counts number of bytes for strings but emojis are represented with some extra bytes in memory so this comparison gives false information.

So it has to be fixed.

### DIFF
--- a/validate.go
+++ b/validate.go
@@ -24,7 +24,7 @@ func MaxLength(length int) Validator {
 	return func(val interface{}) error {
 		if str, ok := val.(string); ok {
 			// if the string is longer than the given value
-			if len(str) > length {
+			if len([]rune(str)) > length {
 				// yell loudly
 				return fmt.Errorf("value is too long. Max length is %v", length)
 			}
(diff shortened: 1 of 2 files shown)


ozguncagri:
Hey @AlecAivazis - $#!t happens in life all the tim, don't worry about it ;)

As you asked I added tests to the code.

AlecAivazis:
Awesome - thanks for doing this. Merging now. 

While I have your attention, do you have any insight into [this issue](https://github.com/AlecAivazis/survey/issues/126)? It's a similar problem that manifests in the custom implementation of the move cursor logic for inputs.

ozguncagri:
Hmm. I wasn't noticed that behavior but I'll check it because I'm using input function for taking byte sensitive data in and it should be fixed. Looks like different alphabets have different behaviors. I have to do some tests I think. If I find something useful I'll send another pull request. Thanks for informing me.

ozguncagri:
Hey @AlecAivazis I made some tests about the issue you pointing at and as I can see ascii and unicode escape sequences are seems like not going to solve it but; I'm just thinking loudly maybe it can be possible reading whole line into slice of rune and removing requested n items from end of it and printing remainin items as string could solve it but I'm not sure about it. I'm not expert of terminal ui's so this whole idea could be nonsense 😅

Issue #130, opened 2018-03-11
Readme screencasts are out of date


AlecAivazis:
With the recent release, the Select/MultiSelect screencasts in the readme are out of date since they do not contain the instructional messages, guiding the user. 

While we're at it, we should add an example of the user filtering the options.

PR #157, opened 2018-10-29
Added multiline input

dpecos:
Basic multiline input. It keeps appending lines to the value until a double empty line is typed.

### DIFF
--- a/multiline_input.go
+++ b/multiline_input.go
@@ -29,7 +29,7 @@ var MultilineInputQuestionTemplate = `
 {{- color "green+hb"}}{{ QuestionIcon }} {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }} {{color "reset"}}
 {{- if .ShowAnswer}}
-  {{- color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
+  {{- "\n"}}{{color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
 {{- else }}
   {{- if .Default}}{{color "white"}}({{.Default}}) {{color "reset"}}{{end}}
   {{- color "cyan"}}[Enter 2 empty lines to finish]{{color "reset"}}
--- a/multiline_input.go
+++ b/multiline_input.go
@@ -30,7 +30,6 @@ var MultilineInputQuestionTemplate = `
 {{- if .ShowAnswer}}
   {{- color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
 {{- else }}
-  {{- if and .Help (not .ShowHelp)}}{{color "cyan"}}[{{ HelpInputRune }} for help]{{color "reset"}} {{end}}
   {{- if .Default}}{{color "white"}}({{.Default}}) {{color "reset"}}{{end}}
   {{- color "cyan"}}[Empty line to finish]{{color "reset"}}
 {{- end}}`
--- a/multiline_input.go
+++ b/multiline_input.go
@@ -32,7 +32,7 @@ var MultilineInputQuestionTemplate = `
   {{- color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
 {{- else }}
   {{- if .Default}}{{color "white"}}({{.Default}}) {{color "reset"}}{{end}}
-  {{- color "cyan"}}[Empty line to finish]{{color "reset"}}
+  {{- color "cyan"}}[Enter 2 empty lines to finish]{{color "reset"}}
 {{- end}}`
 
 func (i *MultilineInput) Prompt() (interface{}, error) {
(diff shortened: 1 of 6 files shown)


MarkusFreitag:
Could you please add tests like the other elements have?

dpecos:
Tests added :D

Just FYI, help interaction like for normal input is not implemented. I've added a test to actually check that.

AlecAivazis:
Hey @dpecos - thanks for adding those tests! Also thanks to @MarkusFreitag for watching over this.

I haven't yet tested this on a windows terminal (I see an ascii escape sequence in the source so my hunch is that it needs to be tweaked a little) so I am going to test it out there before giving it a 👍 and merging it. 

Until I verify it (will hopefully get to tonight or tomorrow) could you add something to the Readme @dpecos  that outlines its behavior? Don't need to worry about the screen recording, I can take care of that (unless you feel up for it).

multiline_input.go
```diff
@@ -0,0 +1,101 @@
...
+		if err != nil {
+			return string(line), err
+		}
+
+		if string(line) == "" {
+			if emptyOnce {
+				numLines := len(multiline) + 2
+				cursor.PreviousLine(numLines)
+				for i := 0; i < numLines; i++ {
+					fmt.Println("\033[K")
```
AlecAivazis: I dont think this will work in a windows environment. This is the ansi escape escape sequence for clearing the rest of the line right? I think you need to update this to use `terminal.EraseLine(i.stdio.Out, terminal.ERASE_LINE_END)`

dpecos: \033 == \x1b (which is what `terminal.EraseLine` is doing) but I agree that using your API is clearer.

I've changed it and adjusted to get the same output. I needed to add an extra \n to make it work the same way.

dpecos: I've added also an example for multiline-input in the README, the only thing missing is the gif (I've used the same one as for input).

Don't forget to create a new one @AlecAivazis !

AlecAivazis:
So I finally got a chance to pull this down a down and give it a try. A few comments:
* It felt like it took one more enter than I thought from the help text. The directions say `Empty line to finish` but i had to leave 2 empty lines. I understand this is to allow a responses with new lines between but it was sort of misleading. Maybe we can change that to `Enter 2 empty lines to finish`
* I was expecting to be able to use my arrow keys to navigate up and edit the lines above - would you mind trying to support that? 

In general, this prompt is a welcome addition to the family. Definitely excited to see this merged since I think it is unique from the rest 💯

AlecAivazis:
Also, once this goes in, i'll start redoing the recordings (which has to happen anyway for #130)

MarkusFreitag:
I also played around with it a bit and actually like the `2 empty lines to finish` thingy. So we should keep it and update the help message.

Regarding the navigation: I would love to see this implemented, but I dont actually know how much effort this takes.

dpecos:
Agree, the message can be misleading, I've changed it to `Enter 2 empty lines to finish`.

About using the cursors to navigate, it would be really nice, but the implementation for that is not trivial at all (at least right now I can't think about an easy way to get that done).

I would suggest keeping current behavior as it is and improve it in a future PR. I would be awesome to have that in my project. BTW, if you want to have a look into it: https://github.com/dpecos/cbox, still a WIP, but planning to release it in a couple of weeks (I still have to work on proper docs and an small tutorial).

AlecAivazis:
Thank you for clarifying the instructional text. I agree that adding the ability to navigate the prompt can be done as a layer on top. I'm going to open an issue when this goes in and if you feel inspired to take it on and want someone to bounce ideas off of - please get in touch (you can find me on the gophers slack channel if you want something more direct than github issues).

I was thinking about this feature and I think the use of the word `Input` is a little vague. Technically, all `Prompts` could be named `...Input` so I've tried avoiding it. To be honest, every time I see the `Input` prompt, I get a little annoyed with my self for picking something so generic. How would you feel if we renamed this to something shorter like `Multiline`?

AlecAivazis:
Your project looks very interesting! I'll definitely be keeping at eye on it as I have wished for something very similar on many occasions.

AlecAivazis:
I spent some time tonight updating all of the screen recordings and got one for this input: https://thumbs.gfycat.com/DisgustingBonyAztecant-size_restricted.gif

Thanks again for the contribution!

dpecos:
I was precisely thinking about this last night :)

I've done the change and adjusted the tests and also renamed the input to just `Multiline`. I really don't mind renaming it.

dpecos:
About cursor navigation, I will try to implement it once I get some time available from cbox, because it's something that will make the app way nicer :)

AlecAivazis:
Awesome! That would be great - I think this is good to go once you update the README to the recording I posted above. It's a little slow for some reason (doesn't look like that in the preview online) but it's better than having something that's not the right prompt.

dpecos:
Done!

skelouse:
> About cursor navigation, I will try to implement it once I get some time available from cbox, because it's something that will make the app way nicer :)

Has this been implemented?

PR #158, opened 2018-11-06
Clear multi-select filter when option is selected

wilhelmeek:
This removes the need to manually clear the filter when you want to change the input after selecting an option in multi-select.

### DIFF
--- a/multiselect.go
+++ b/multiselect.go
@@ -27,7 +27,6 @@ type MultiSelect struct {
 	Help          string
 	PageSize      int
 	VimMode       bool
-	ClearOnSelect bool
 	FilterMessage string
 	filter        string
 	selectedIndex int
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Hey @wilhelmeek - thanks for implementing this!

I know its a small change, but would you mind adding a test to verify this behavior?

wilhelmeek:
@AlecAivazis for sure!

wilhelmeek:
@AlecAivazis I've added a prompt test that tests this behaviour. Not sure if this is would be comprehensive enough to assert this behaviour, so please let me know if you have alternative suggestions 😄

AlecAivazis:
Thanks @wilhelmeek!

I think its enough because if it didn't clear the filter, Tuesday would have stayed highlighted 👍

Issue #160, opened 2018-11-08
Add a message prompt feature


rmfish:
Add a info/warning message prompt without any answer.

rmfish:
In my case, i use the survey ask one by one,  it always do something and send feedback between the asks. So i want  a message prompt without answer to display the feedback :)

coryb:
I am not sure I understand your use-case, but prompts are just a Go interface.  You can provide your own implementation in your code.  If you don't want to re-render the prompt with the answer you could just implement a no-op `Cleanup` function:
```
type MyConfirm struct {
    Confirm
}

func (my *MyConfirm) Cleanup(val interface{}) error {
    return nil
}
```

PR #161, opened 2018-11-08
Add text prompt feature

rmfish:
Add a text prompt prints message without any answer.
#160

### DIFF
--- a/confirm.go
+++ b/confirm.go
@@ -15,6 +15,10 @@ type Confirm struct {
 	Help    string
 }
 
+func (*Confirm) NeedAnswer() bool {
+	return true
+}
+
 // data available to the templates when processing
 type ConfirmTemplateData struct {
 	Confirm
--- a/core/template.go
+++ b/core/template.go
@@ -31,6 +31,9 @@ var (
 	// SelectFocusIcon is prepended to an option to signify the user is
 	// currently focusing that option
 	SelectFocusIcon = ">"
+
+	// TextIcon will be be shown before an text
+	TextIcon = "!"
 )
 
 /*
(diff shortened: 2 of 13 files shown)


AlecAivazis:
Hey @rmfish  - thanks for submitting this.

I'm a bit uneasy with changing the interface for `Prompt` since it would be an API breaking change for survey. Given that you can pull this off with `AskOne` or other hacky approaches like printing from `Transform`, i'm inclined to think that this doesn't add enough to the collection of default prompts and is better kept in userland. 

For your usecase, it seems very likely that the message you want to show the user depends on the answers to the other prompts which means `Text` either needs to take a function of the previous values to define its message, or it would need to be created in an `AskOne` anyway so that `Message` can refer to the struct holding the answers.

AlecAivazis:
Hey @rmfish i appreciate you also adding support for a go module. Would yu mind opening that up in a second PR so I can merge that bit ASAP?

rmfish:
> Hey @rmfish i appreciate you also adding support for a go module. Would yu mind opening that up in a second PR so I can merge that bit ASAP?

OK :)

rmfish:
> Hey @rmfish - thanks for submitting this.
> 
> I'm a bit uneasy with changing the interface for `Prompt` since it would be an API breaking change for survey. Given that you can pull this off with `AskOne` or other hacky approaches like printing from `Transform`, i'm inclined to think that this doesn't add enough to the collection of default prompts and is better kept in userland.
> 
> For your usecase, it seems very likely that the message you want to show the user depends on the answers to the other prompts which means `Text` either needs to take a function of the previous values to define its message, or it would need to be created in an `AskOne` anyway so that `Message` can refer to the struct holding the answers.

Sorry, I was not realized that it's bad to change the api of  `Prompt`.  There are indeed other better ways to resolve my problems. I close this PR, thx.

PR #162, opened 2018-11-09
support go module

rmfish:


go.mod
```diff
@@ -0,0 +1,16 @@
+module gopkg.in/AlecAivazis/survey.v1
```
AlecAivazis: Do we still need the `v1` in our imports since this can be considered the first version of the package?

Reading [the page on the go repo](https://github.com/golang/go/wiki/Modules) makes it sound like you only have to have the version in the import path for major versions > 1

rmfish: I think it's required.  
Because the major version `v1` is contains in the import path: `import gopkg.in/AlecAivazis/survey.v1`.

It can use `module github.com/AlecAivazis/survey` without version, but user should use `replace` in go.mod
```
import gopkg.in/AlecAivazis/survey.v1 v1.6.4
replace gopkg.in/AlecAivazis/survey.v1 => github.com/AlecAivazis/survey v1.6.4
```

or direct import `github.com/AlecAivazis/survey` in gofile and go.mod
```
import github.com/AlecAivazis/survey v1.6.4
```

AlecAivazis: Yea i guess i was suggesting that we remove it from the canonical import path. This wouldn't break people's current imports since gopkg.in would still resolve to the v1 releases

rmfish: I think there is an another problem:

Some golang files in survey project also use `import gopkg.in/AlecAivazis/survey.v1`, such as [core/renderer.go](https://github.com/AlecAivazis/survey/blob/master/core/renderer.go)

if we use `module gopkg.in/AlecAivazis/survey.v1`,these files will build failed.
In contrast, if we change these gofile's import path to `github.com/AlecAivazis/survey`, I guess the project used gopkg will get failure in run time

AlecAivazis: After talking it over with @coryb i think the right move here is to go to `v2` with the `go.mod` file. This will keep everyone's imports working since gopkg.in will still resolve v1 to a version of the source with the `v1` references

AlecAivazis:
Thought about this some more... since it would put a hard divide in the usage, i'm going to hold off on migrating `survey` to modules until there is a good reason to do so. I am wary of bumping a major version for nothing more than an import change. 

Sorry for stringing this PR along. It was a good exercise for me to think through and it just doesn't seem like the right time for the package.

PR #165, opened 2018-11-29
Prompt again

orymate:
Recall the previous invalid input in the repeated Editor prompt so the user doesn't have to start over typing a long answer (for example a json structure).

Add optional PromptAgain method to Prompts to back this.

### DIFF
--- a/editor.go
+++ b/editor.go
@@ -72,7 +72,20 @@ func init() {
 	}
 }
 
+func (e *Editor) PromptAgain(invalid interface{}, err error) (interface{}, error) {
+	initialValue := invalid.(string)
+	return e.prompt(initialValue)
+}
+
 func (e *Editor) Prompt() (interface{}, error) {
+	initialValue := ""
+	if e.Default != "" && e.AppendDefault {
+		initialValue = e.Default
+	}
+	return e.prompt(initialValue)
+}
+
+func (e *Editor) prompt(initialValue string) (interface{}, error) {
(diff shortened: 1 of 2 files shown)


orymate:
Hello @AlecAivazis, could you please check this PR?

AlecAivazis:
Hey @orymate - sorry for lagging a bit on this. 

I think this makes sense. Forcing the user to re-enter something they were just working on seems like a bad user experience.

Overall, the code looks fine. Would you mind adding a test for this?

orymate:
Hello,
does the test in 75e46df make sense? I did not catch the intention behind the go routine I removed, but it triggered errors after the extension, while it did work for me sequentially.

AlecAivazis:
Thanks for adding that - it looks like you changed an existing test to cover your needs which makes it hard to track test failures (since the test is really testing 2 different things). Could you move the test into another test case?

orymate:
Added separate test.

orymate:
@AlecAivazis, can I do anything more to have this merged?

AlecAivazis:
Hey @orymate!

I tried to add a test that would showcase the behavior to `tests/editor.go` but I couldn't get it to work. Could you add one there so I can run this on my laptop?

orymate:
@AlecAivazis, I added the manual test code

AlecAivazis:
Awesome! Thanks for adding that - worked like a charm.

PR #167, opened 2018-12-16
Use space to select option

johnksv:
The default behaviour for a `select` is that space will append  " " to the filter.  This PR adds support to use space to select an option instead of appending to the filter.

Why? in my use case it doesn't make sense to filter on/with space. Since space is used to select/deselect in `multiselect` I also think it makes sense that it can be used select an option from an `select`. 

This PR is not related to any issues, so please feel free to do what you would like with the PR.


johnksv:
Closing this one. See you points. Feel free to use the code if you change your mind in the future. :)

PR #171, opened 2019-01-10
Feature: allow custom filtering behavior for Select and MultiSelect

metacosm:
Would allow for more complex filtering than simple substring matching (e.g. based on extra metadata in the case options are complex types instead of simple strings).

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,29 @@ change the global `survey.PageSize`, or set the `PageSize` field on the prompt:
 prompt := &survey.MultiSelect{..., PageSize: 10}
 ```
 
+### Custom filtering in Select and MultiSelect
+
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -188,6 +188,29 @@ change the global `survey.PageSize`, or set the `PageSize` field on the prompt:
 prompt := &survey.MultiSelect{..., PageSize: 10}
 ```
 
+### Custom filtering in Select and MultiSelect
```
AlecAivazis: Thanks for adding documentation!

AlecAivazis: Since this is something that is shared across the select and multi select prompts, should we add something to the table of contents?

metacosm:
Thanks!

PR #173, opened 2019-01-12
Fix validation example code

bmuschko:
Function needs to return a value for error. Add missing comma.


AlecAivazis:
Thanks for fixing this!

PR #175, opened 2019-01-19
Removed extra empty line in render for Multiline

dpecos:
Small fix for Multiline input that was rendering an extra empty line compared with the rest of input types.

Before:
```
? Code / Command

Using previous value: 'code'
```

After:
```
? Code / Command
Using previous value: 'code'
```

### DIFF
--- a/multiline.go
+++ b/multiline.go
@@ -29,7 +29,8 @@ var MultilineQuestionTemplate = `
 {{- color "green+hb"}}{{ QuestionIcon }} {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }} {{color "reset"}}
 {{- if .ShowAnswer}}
-  {{- "\n"}}{{color "cyan"}}{{.Answer}}{{color "reset"}}{{"\n"}}
+  {{- "\n"}}{{color "cyan"}}{{.Answer}}{{color "reset"}}
+  {{- if .Answer }}{{ "\n" }}{{ end }}
 {{- else }}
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Hey @dpecos thanks for another contribution! Can you post the prompt that you are using? I can't seem to replicate the empty line on my laptop

dpecos:
Happy to help as much as possible! This lib has been really useful for my projects.

The code I'm using is something like this (I've removed/changed non interesting code for this issue):

```
label := "Code / Command"
help := "Blank entry keeps previous value. Ctrl+C clears it"

prompt = &survey.Multiline{
    Message: label,
    Help:    help,
}

var value string

survey.AskOne(prompt, &value, nil)

fmt.Println("Using previous value: 'code'")
```

So once entered the multiline value, the rendered result of the input generates an extra empty line just before the last `fmt.Println()`.

Let me know if this explains the issue well enough, otherwise I'll try to record something.

dpecos:
The issue only happens if the input is an empty string (in the case of multiline, just type twice \n).

Once the value is rendered back to the input, just in this case when the value entered is empty, an extra new line results in the render.

In my example, the ouput `Using previous value: 'code'` comes from an extra fmt.Println, not from the input.

Let me know if this clarifies the issue, otherwise I'll create an small program for it.

AlecAivazis:
Ah thanks for the explanation! I also had a hard time reproducing this. I think this is ready to go so i'm going to merge it now. 

Thanks for the contribution @dpecos!

dpecos:
Thank you both!

PR #177, opened 2019-01-25
Add module enabled build support

chappjc:
This creates the go.{mod,sum} files, which define survey as a module and lock the dependencies.

`go mod init` was used to convert the previous dep files and keep the same dependencies versions.

This should allow projects that use go modules to import and build survey with the intended versions of its dependencies.


AlecAivazis:
Awesome. Thanks for doing this. I've gone back and forth on wether this is worth it or not but it doesn't really change much if we leave the import path still referencing `gopkg`. With the next API change we'll take the opportunity to update the path to `github.com/AlecAivazis/survey`

chappjc:
No problem.  However, I think that because of gopkg.in, which still points to 1.8.1, it looks like a new tag is needed to allow the go.mod files to use by other projects that import survey.  Thanks!

chappjc:
Would it be possible to tag 1.8.2 for the above reason?  I appreciate it!

MarkusFreitag:
Awesome, thanks for adding the go modules support! :+1:

PR #180, opened 2019-02-28
fix(cursor): fix terminal cursor move display

mritd:
fix the problem of moving the cursor display when the terminal inputs non-English characters (such as Chinese, Japanese)

refs AlecAivazis/survey#126

Signed-off-by: mritd <mritd1234@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/kr/pty v1.1.1 // indirect
 	github.com/mattn/go-colorable v0.0.9 // indirect
 	github.com/mattn/go-isatty v0.0.3
+	github.com/mattn/go-runewidth v0.0.4
 	github.com/mgutz/ansi v0.0.0-20170206155736-9520e82c474b
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/stretchr/testify v1.2.1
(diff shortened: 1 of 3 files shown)


mritd:
I fixed the cursor display problem using the https://github.com/mattn/go-runewidth library, 
but I don't know how to write a complete test case to test it... 

I did a lot of testing locally, now it looks good work 😂

AlecAivazis:
Thanks for submitting this @mritd. 

I tried using 👫as an input and got some weird behaviors if I moved around and tried to delete it. I haven't yet taken the time to investigate if this is a problem with the `go-runewidth` or something in here but I thought I would raise it to you.

mritd:
@AlecAivazis Have you encountered a problem with the cursor suddenly appearing on the next line?

mritd:
I found in the test: When pressing the left arrow (move button) and the delete button at the same time, there is a certain probability that the cursor will suddenly appear on the next line.



I was testing the master branch code (not merging this PR); I guess this may be due to a location variable thread is not safe.

AlecAivazis:
@mritd any luck hunting this down? There are a bunch of changes that have been merged recently so you probably want to open a new PR based on the current master

PR #185, opened 2019-03-27
fix: lets add help about selecting multi-selects using space

jstrachan:
as folks sometimes get confused how to select items in a multi-select

### DIFF
--- a/multiselect.go
+++ b/multiselect.go
@@ -52,7 +52,7 @@ var MultiSelectQuestionTemplate = `
 {{- color "default+hb"}}{{ .Message }}{{ .FilterMessage }}{{color "reset"}}
 {{- if .ShowAnswer}}{{color "cyan"}} {{.Answer}}{{color "reset"}}{{"\n"}}
 {{- else }}
-	{{- "  "}}{{- color "cyan"}}[Use arrows to move, type to filter{{- if and .Help (not .ShowHelp)}}, {{ HelpInputRune }} for more help{{end}}]{{color "reset"}}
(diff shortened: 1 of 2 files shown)


jstrachan:
@AlecAivazis I don't have karma to retry the CI build but I don't see any actual errors ;)

jstrachan:
@AlecAivazis actually no the `run tests` does fail if we modify the help text in any way in `multiselect*.go` - I've no clue why that causes a failure though! I'm a bit stumped really - any ideas? I can't seem to make any modification to the template in `mulitselect.go` if I make the same change in `multiselect_test.go` - the result is always that the test case hangs and then fails

AlecAivazis:
Hey @jstrachan! thanks for another contribution. Unfortunately, i'll be on vacation until Saturday so I won't be able to take a look until I get back.

jstrachan:
no worries! I'm totally stumped why I can't change this label in any way (e.g. changing 1 character fails the tests) so was hoping you had a clue ;)

jstrachan:
I figure we could change multi-select + select help to specify how to select (spacebar for multi-select + return for select?) basically. If anything folks guess they can use the arrow keys, but folks often try return for multi-select instead of spacebar.

Maybe return on a multi-select with mandatory values + no selection could show a hint like `please select an item first with the spacebar` or something?

AlecAivazis:
Very strange. I'm also seeing the same behavior. I'm really not sure what is causing it. It could be a problem with go-expect. Maybe @coryb or @hinshun has an idea?

AlecAivazis:
Okay I finally got a chance to give this a try on another machine. I was not able to run the tests on my local copy of your branch however, I made the same changes to master and the tests ran. So I'm going to close this and open a PR with the same changes so we can get them in.

PR #204, opened 2019-05-30
runs test on osx and windows

coryb:
Travis now supports windows in beta mode, the build setup is a bit slow, but it ran for me in a few minutes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: go
 
+os:
+  - linux
+  - osx
+  - windows
+
 go_import_path: gopkg.in/AlecAivazis/survey.v1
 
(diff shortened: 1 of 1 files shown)


coryb:
I am not entirely sure what environment the windows tests run under.  The docs state it is using git-bash by default: https://docs.travis-ci.com/user/reference/windows/#git-bash 
I am not sure what sort of tty environment is setup via git-bash. 

I was trying to figure out how to run the tests via powershell, but I don't understand how we would run powershell for just the windows build.

AlecAivazis:
Oh cool! Verifying gitbash is better than nothing. Thanks for tweaking this

PR #211, opened 2019-05-31
Fixed behavior of Select when all answers are filtered out

AlecAivazis:
This PR fixes the behavior of Select to not return an empty string if there is a filter applied that clears the available options.

This PR will have to get merged after #209 and #208 since it's based on a branch where they were merged.

fixes #208


coryb:
I will review this one after the other 2 prs are merged, it is hard to see what is changing for this PR.

AlecAivazis:
This branch got a little messed up. I'm going to open another with just the commits that matter

Issue #201, opened 2019-05-28
Remove config via global variable


AlecAivazis:
Branching off of the conversation in #192, we should see if we can use variadic configuration functions to remove as much configuration via global variables as possible.

The things we should consider removing from global config:
- [x] Validators to `AskOne`
- [x] Icon Text
- [x] Help input rune
- [x] PageSize 
- [x] Default filter

ahmetb:
+PageSize

AlecAivazis:
@ahmetb thanks for the suggestion! I'm curious if that was because you think there is a better default? I pulled 7 out of nowhere so if you have any reason for a particular number, i'd love to hear it

ahmetb:
I don't need a better default. :) I just need to configure it per-question.

AlecAivazis:
Well you can already configure it per prompt by setting the `PageSize` field on the `Select` and `MultiSelect` structs but I'm adding the ability to set it with an `AskOpt` now

Issue #105, opened 2017-10-20
Support for display name in Select prompts 


chzzzh:

AlecAivazis:
Hey @frodgesh! Thanks for opening this issue. Do you mind adding some more information as to why you think the API should be changed this way? since it's a breaking change, id like to think through this throughly before changing anything. 

Currently, select and multi select return the value the user chose so that options can themselves be ints. Also, if the selects return ints, we'll have to add of logic to prevent the casting of the return value since it doesn't make a lot of sense to convert the index into a string.

frodgesh:
@AlecAivazis, thanks for the feedback! Of course, user can convert string answer to int, but that's not the point. If we send []string to the quiz, certainly we do have the array, so why do we have to get string as a return? Would be much likely to get an index of this string to find it in the array, I guess.

In example, we got a struct array of food with two fields: name and calories. We'll send to the quiz a string[] copy of this array, but how do we know, how much calories in chosen food if we get a string? Probably, indexing the whole array would not be the safer idea.

AlecAivazis:
So I thought about this last night, I think the right approach would be something very similar to the `settable` interface which could allow for you to pass your food structs as questions. Then, when a user selects the option, the struct associated with their selection would be written to the field. What do you think?

frodgesh:
@AlecAivazis It is even more acceptable based on the idea that a function should return the answer, not a reference.

thadeetrompetter:
Hi! In my case, i'm feeding a slice of string 'questions' into the select, which may contain duplicates. In this situation, the returned string does not really help to get a reference to the right question metadata that i'm also storing. A returned int, as well as being able to pass structs as options, would be of great help.

gregoriokusowski:
Hey, I made a few changes to make it work with the `Stringer` interface.
It's still WIP and is not working with multiselect yet.

I can open a proper PR if you guys think that this is a good approach.
https://github.com/AlecAivazis/survey/compare/master...gregoriokusowski:wip-select?expand=1

The idea basically is to keep the current API, creating the `option` Stringer for old string options.
Instead of returning the index, we can return (or copy) your object back.
WDYT?

AlecAivazis:
Thanks for working on this @gregoriokusowski! I have been meaning to start implementing this but have gotten completely swamped with getting some work out for a client as quickly as possible. 

[...]

PR #214, opened 2019-05-31
Added WithPageSize AskOpt

AlecAivazis:
This PR is part of the work summarized in #201. It adds the `survey.WithPageSize` `AskOpt` that replaces the modification of `survey.PageSize` as the way to modify the default page size for prompts.

Since we now have to have a place for "global" configuration of prompts, I had to change the `Prompt` interface to include a reference to the config.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,10 +159,16 @@ The user can also press `esc` to toggle the ability cycle through the options wi
 
 By default, the select prompt is limited to showing 7 options at a time
 and will paginate lists of options longer than that. To increase, you can either
-change the global `survey.PageSize`, or set the `PageSize` field on the prompt:
+set the `PageSize` field on the prompt:
 
 ```golang
 prompt := &survey.Select{..., PageSize: 10}
+```
+
+Or pass an an `AskOpt` to `survey.Ask` or `survey.AskOne`:
+
+```golang
+survey.AskOne(prompt, &days, survey.WithPageSize(10))
 ```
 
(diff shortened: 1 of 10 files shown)


AlecAivazis:
Yea I thought about that while I was putting this together but it felt like a lot of ceremony for an if statement, especially considering I'd still have to pass the `config` in. There could be an argument for encapsulating the `paginate` function into a method. I'll take a look and see if its worth it before merging - thanks for the review!

AlecAivazis:
So I tried to consolidate the `paginate` logic into a method on a `Pager` interface like you suggested. Unfortunately, since you can't set promoted fields, this would force the API to be:

```golang
Select{
    Pager{ PageSize: 7 }
}
```

which is just a little too verbose for me

coryb:
Okay, yeah, I agree, forgot about that aspect.  We could minimize this issue by having a `NewSelect` etc, with variadic args, ie `NewSelect(opts ...func(*Select))` and `WithPageSize(s *Select)`. Not sure if there are other issues that would be resolved or made worse by this direction though.

AlecAivazis:
Yea that's something to keep in mind - it starts to become very verbose and I think so long as we have sane interpretations of a zero value, we should keep the struct-based API as I think its much simpler for people to grok quickly

AlecAivazis:
It just occurred to me that it's also very possible that sort of API is what a solution to #105 would look like.

Issue #141, opened 2018-05-10
Add support for changing the color of the different runes/prompts


kleoken:
Creating a CLI with this currently and would be nice if I could change the color of the QuestionIcon application wide

AlecAivazis:
Hey @Celo44 - thanks for submitting this. I don't know if I will have the time in the near future to add this but if you wanted to give it a stab, I would be happy to review a PR 🙂

MarkusFreitag:
I would start by making the icons configurable. Therefore I would like to evolve to icon, which is currently a `string`, into a `struct` which for now has two fields.
```
type Icon struct {
  Symbol string
  Color    string
}
```
Maybe @AlecAivazis can give some feedback about this approach. If it is accepted I will provide a PR.

AlecAivazis:
Hey @MarkusFreitag - thanks for the input. I like the idea, although I'm not sure if providing this would be too much customizability for the library. Can you maybe share your reason for wanting to change the colors? 

In general, i'm a bit wary of changing the public API for a relatively minor feature. I think I would prefer if there was a way to call something like `surveyCore.SetIcon("ErrorIcon", Icon{Symbol: "!"}) which keeps the top level API the same. In this case, we would add variables for each of the colors, and `SetIcon` would change the values of `ErrorIcon` to the `Symbol` field and the color templates to match that field if provided.

Thoughts? Would that satisfy both of your needs?

MarkusFreitag:
Thanks for your feedback. It does not really changes the API that much. Basically I have just moved the hard-coded color definitions from the template strings into the icon struct and then call `Icon.Color` in the template. Also the `var TemplateFuncs = map[string]interface{}` works as before. I think that it also makes it easier to maintain the library, as you have such things in a central place and not spread over every template. Nevertheless I can push my changes into a branch, which might makes it easier to find a consent on this.

> Can you maybe share your reason for wanting to change the colors?

Passing this question to @Celo44

MarkusFreitag:
@AlecAivazis I put the changes into a separate branch, maybe in the next days you have some time to have a look on them. As written before the API behaviour does not change.
https://github.com/MarkusFreitag/survey/commit/ceb60e1791ab3d800b1fe5957f246e3093900412

bmuschko:
@AlecAivazis Any chance we can move forward with @MarkusFreitag's commit?

AlecAivazis:
[...]

PR #216, opened 2019-06-01
Added missing AskOpts

AlecAivazis:
This PR continues the work in #201  and adds the remaining `AskOpts`:

* `survey.WithHelpInput` to change the rune that prompts look for to provide help.
* `survey.WithIconSet`  to change the icons
* `survey.WithFilter` to change the default filter applied to Select/MultiSelect

This has to go in after #214 since it relies on the `PromptConfig` which was introduced it that PR.


core/renderer.go
```diff
@@ -35,13 +40,17 @@ func (r *Renderer) NewCursor() *terminal.Cursor {
 	}
 }
 
-func (r *Renderer) Error(invalid error) error {
+func (r *Renderer) Error(invalid error, icon string) error {
```
coryb: For consistency, I think we should pass in the IconSet here.  Then the template would look like `{{.Icons.Error}}`. I realize that only one Icon is used for now, but perhaps we will add another icon eventually or a user will want to make a more elaborate ErrorTemplate that utilizes other icons?

AlecAivazis: Yea I had that initially but then moved the `IconSet` definition to `survey` so that `WithIconSet`s signature looked reasonable which created a circular import since we render the Error template in `core`. Maybe the solution is to move `Renderer` into the root package?

AlecAivazis:
@coryb there is now parity across the methods in the Prompt interface. Not sure why I wanted to avoid another file in the root of the project when there are already so many...

AlecAivazis:
I thought a bit more about this API and i think one of two things needs to happen. Either I should remove '`WithHelpInput`  or `HelpInput` should be removed from the `IconSet` struct. It's not clear from the API what would happen if you passed both `WithHelpInput` and a `WithIconSet` that modified the icon since they modify the same field in the icon set.
 
Any preference @coryb? Keeping it in the `IconSet` has a smaller API surface since its kind of tucked away. But it's also not really an "icon" like the others since it's input and would never include any kind of color configuration regardless of how #141 shakes out.

PR #220, opened 2019-06-03
editor: add FileNamePattern option

orymate:
You may want to specify the file name pattern for the temporary file
created to activate the proper syntax highlighting settings in the
user's editor automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,9 +233,15 @@ survey.AskOne(prompt, &days, survey.WithPageSize(10))
 
 ### Editor
 
-Launches the user's preferred editor (defined by the \$EDITOR environment variable) on a
+Launches the user's preferred editor (defined by the \$VISUAL or \$EDITOR environment variables) on a
 temporary file. Once the user exits their editor, the contents of the temporary file are read in as
 the result. If neither of those are present, notepad (on Windows) or vim (Linux or Mac) is used.
+
+You may want to specify the file name pattern for the temporary file to activate syntax highlighting in the editor.
+
+```golang
+survey.AskOne(&survey.Editor{Message: "Shell code snippet", FileNamePattern: "*.sh"}, &content)
+```
 
 ## Filtering Options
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -202,10 +202,16 @@ survey.AskOne(prompt, &days, survey.WithPageSize(10))
...
 
-Launches the user's preferred editor (defined by the \$EDITOR environment variable) on a
+Launches the user's preferred editor (defined by the \$VISUAL or \$EDITOR environment variables) on a
 temporary file. Once the user exits their editor, the contents of the temporary file are read in as
 the result. If neither of those are present, notepad (on Windows) or vim (Linux or Mac) is used.
 
+You may want to specify the file name pattern for the temporary file to activate syntax highlighting in the editor.
+
+```golang
+survey.AskOne(&survey.Editor{Message: "Shell code snippet", FileNamePattern: "*.sh"}, &content)
```
AlecAivazis: I think we can do away with the `Pattern`. In fact, I think the simpler API is to name it `Extension` and skip the `*` bits

AlecAivazis:
Thanks for the contribution @orymate!

In general, I think the feature makes a lot of sense. Looking forward to getting this merged.

README.md
orymate: I considered the same, but I had some (maybe rare) use cases in mind when a prefix might be important too. It may help communicating to the user something like which file (answer) are they editing, like when you have to provide two different snippets, and might forget which one are you currently editing. Actually, the `Extension` version is good for all my current use cases, so it's ok for me as well.

AlecAivazis: Hm, yea that's a good point. It's not like it imposes a lot of overhead so keeping it could be the right balance. My biggest concern would be that we're responsible for implementing that pattern logic but I see that `ioutil` provides it out of the box so that's not a problem. Let's keep it.

Maybe add a link to the documentation that it follows the same format as [this function](https://golang.org/pkg/io/ioutil/#TempFile)

AlecAivazis:
Thanks!

PR #223, opened 2019-06-05
Add MultiInput prompt

eimlav:
This commit adds a MultiInput prompt allowing a user to enter multiple
distinct values for a question. These are then returned as a []string.
For example:

	? What are your friend's names?

Then the user is prompted for the first answer, then the second and so
on.

	? #1: Batman
	? #2: Joker
	? #3:

Then once finished they enter an empty line and the question will exit,
outputting the answers.

	? What are your friend's names? [Batman Joker]

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,7 @@ for the old `v1` version, see [here](https://godoc.org/gopkg.in/AlecAivazis/surv
 1. [Examples](#examples)
 1. [Prompts](#prompts)
    1. [Input](#input)
+   1. [MultiInput](#multiinput)
    1. [Multiline](#multiline)
    1. [Password](#password)
    1. [Confirm](#confirm)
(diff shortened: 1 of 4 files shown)


eimlav:
@AlecAivazis I was unable to get [`TestMultiInputPrompt`](https://github.com/AlecAivazis/survey/blob/7f9c8397fea902ccc72d583a26a0dfc52a2df04c/multiinput_test.go) tests to work. Would you be able to provide some advice on how I could get these working?

AlecAivazis:
Hey @eimlav - thanks for submitting this.

We've had a few discussions over the years about a prompt like this and ultimately we decided that this a prompt like this is too similar to `Input` to add to the group. That's not to say its not a useful prompt. The prompts in survey are designed as building blocks for prompts like this one so we try to only add ones that are truly unique.

Sorry about this! I know it can be disappointing to put a bunch of work into something that won't get accepted.

eimlav:
@AlecAivazis Ah I understand. I'll just make it a part of the package I'm using it in then. At least this saves me from having to write those tests! Thanks for the quick response.

PR #225, opened 2019-06-13
Select index

### LINKED ISSUE #105: Support for display name in Select prompts

AlecAivazis:
This PR tweaks the values written by `Select` to be either the option's index or the value, based on the type of the target field. I also had to change the type of the `Default` field in both prompts to `interface{}` to accommodate a user who is thinking in terms of either indices or values.

The old implementation of MultiSelect had an issue when the string values were not unique. In order to address that, I changed the internal tracking of the selected options to a map keys on index. This meant that all of the paginating logic had to hold onto some information that tracked the original index. In order to make this all work nicely, I changed the signature of the `Filter` function passed around to be `func (filterValue string, optValue string, optIndex int) bool`. This saves the user from writing the boilerplate that loops over their lists and applies the same function. It also has the extra benefit of allowing us to track which indices passed without forcing the user to provide that information.

fixes #105


README.md
```diff
@@ -250,15 +258,9 @@ is active. This will filter out all options that don't contain the typed string
...
-    for _, v := range result {
-        if len(v) >= 5 {
-            filtered = append(filtered, v)
-        }
-    }
-
-    return filtered
+func myFilter(filterValue string, optValue string, optIndex int) include bool {
+    // only include the option if it includes the filter and has length greater than 5
+    return strings.Contains(optValue, filter) && len(optValue) >= 5
```
coryb: the variable should be `filterValue` I think rather than `filter`.

PR #227, opened 2019-06-19
Fix(doc): replace function by func in code snippet

seblegall:
Hi,

I've spotted a little mistake in the code snippet about changing the survey icon. Here is the fix. 

Cheers,


AlecAivazis:
Whoops! Thanks for fixing this 👍

PR #240, opened 2019-08-01
fix(core): Allow to use `time.Duration` struct fields

liranp:
## Description

We are trying to use this package to initialize a `time.Duration` field but getting an error when the answer is written to the target struct by [core/write.go#L189-L190](https://github.com/AlecAivazis/survey/blob/0f64ceaa253b1069b6c235e74e056213e7a567e2/core/write.go#L189-L190). This PR adds support for struct fields of type `time.Duration`.

## Example

```go
type Options struct {
	Timeout time.Duration
}

survey.Ask([]*survey.Question{{...}}, new(Options))
```

* Answering with a string in an integer format (for example, **30**) returns the following error: `reflect.Set: value of type int64 is not assignable to type time.Duration`

* Answering with a string in a valid duration format (for example, **30s**) returns the following error: `strconv.ParseInt: parsing "30s": invalid syntax`

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -6,6 +6,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"time"
 )
 
 // the tag used to denote the name of the question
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Nice! Looks like we have an issue in travis that's preventing your build from passing. I'll try to figure out what's going on so we can get this merged ASAP

AlecAivazis:
Okay so the latest PR should fix your build problem. Mind rebasing on the latest master?

liranp:
Done. Thank you!

Issue #241, opened 2019-08-02
v2 version from the code should be removed


mkumatag:
The specific version(e.g: github.com/AlecAivazis/survey/v2/core, github.com/AlecAivazis/survey/v2/terminal) in the pkg import path usually used when the package is intentionally referring to some other version(usually used for importing the features from the previous version for the special use case like package backward compatibility etc..).

In survey usecase, I don't see a point of having a version in the import path, hence need to be removed from all the root and test files.

AlecAivazis:
I appreciate your post. We added the `v2` to the path when we broke the API. I'm going to keep it as it is for now so that we don't break people's worlds

mkumatag:
> I appreciate your post. We added the `v2` to the path when we broke the API. I'm going to keep it as it is for now so that we don't break people's worlds

Well, this is creating an issue while vendoring the code via `dep` tool etc.. The current v2 format can be vendored only via `go mod` which is not widely used yet.

This impacts all the user who is using package and trying to bump the version via `dep` tool.

e.g:
```shell
$ pwd
/Users/manjunath/survey_ws/src/github.com/mkumatag

$ more example.go
package main

import (
        "fmt"
        "github.com/AlecAivazis/survey/v2"
)

// the questions to ask
var qs = []*survey.Question{
....
.....

$ dep init
init failed: unable to solve the dependency graph: Solving failure: No versions of github.com/AlecAivazis/survey met constraints:
	v2.0.2: Could not introduce github.com/AlecAivazis/survey@v2.0.2, as its subpackage github.com/AlecAivazis/survey/v2 is missing. (Package is required by (root).)
	v2.0.1: Could not introduce github.com/AlecAivazis/survey@v2.0.1, as its subpackage github.com/AlecAivazis/survey/v2 is missing. (Package is required by (root).)
	v2.0.0: Could not introduce github.com/AlecAivazis/survey@v2.0.0, as its subpackage github.com/AlecAivazis/survey/v2 is missing. (Package is required by (root).)
	v1.8.5: Could not introduce github.com/AlecAivazis/survey@v1.8.5, as its subpackage github.com/AlecAivazis/survey/v2 is missing. (Package is required by (root).)
	v1.8.4: Could not introduce github.com/AlecAivazis/survey@v1.8.4, as its subpackage github.com/AlecAivazis/survey/v2 is missing. (Package is required by (root).)
[...]

PR #242, opened 2019-08-02
Remove v2 versioning

mkumatag:
Fixes: #241

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ os:
   - osx
   - windows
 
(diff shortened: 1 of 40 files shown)


AlecAivazis:
As I said in #241, we're not going to remove the `v2` in the import path since it would break user's bilds

tsingson:
> As I said in #241, we're not going to remove the `v2` in the import path since it would break user's bilds

yes, don't do this.   we use v2 in product for half year............. it's work great.

PR #239, opened 2019-07-31
Add support for ppc64le architecture

mkumatag:
As the title mentioned this PR is to add support for ppc64le architecture

PR #243, opened 2019-08-13
survey: prevent panic on nil AskOpt

orymate:
nils may remain from v1 code, and not trivial to find. Also, dynamic
generation of optional requirements is more complex than needed this
way:

```golang
opts := []survey.AskOpt{}
if required {
    opts = append(opts, survey.WithValidator(survey.Required))
}
survey.AskOne(input, &output, opts...)
```


AlecAivazis:
Thanks for the contribution! I can see how this would be useful for migrating from v1 -> v2 :+1:

AlecAivazis:
I'm not sure why the build is failing. I thought this was fixed in #239

PR #248, opened 2019-08-24
style(README): Ensure first example in README.md imports from a path that works

CaffeinateOften:
Fix version typo in main readme example

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ package main
 
 import (
     "fmt"
-    "github.com/AlecAivazis/survey/v2"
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Thanks for the contribution. This import is in fact not an error is because of the way go handles versioning modules. You should see a consistent path everywhere in the project

CaffeinateOften:
Ah thanks. Reviewed go module documentation, set up dir outside of GOPATH, init'ed my own go.mod file, and added explicit require statement within go.mod to get things working correctly.

PR #260, opened 2019-11-20
Change selection prompt from space to enter.

patrickdillon:
When using select or multiselect, the user is prompted to make a selection using space, but entering space causes results to be filtered by space. This commit updates the prompt to tell users to make a selection using the enter key.


AlecAivazis:
Ah nice catch! And thanks for your first submission! 

I'll try restarting the failed build to see if that gives us the green light to merge. Since this is mostly a copy change it should be very quick to get in.

sdodson:
@AlecAivazis Can you take another look at this?

AlecAivazis:
Yep! Sorry that took so long

AlecAivazis:
This should now be available in `v2.0.5`

patrickdillon:
@AlecAivazis can we cherry-pick this to v1? We have plans to upgrade to v2 but it is on the roadmap. If there is a process for cherry-picking to v1 that I could help with let me know.

PR #263, opened 2019-12-25
Add core.IsFieldNotMatch function.

kataras:
Hello @AlecAivazis,

I am pushing this PR in order to check if an error is caused of an unmatched field. There are cases when a map can be ignored (e.g. when you have a Transform which stores a value to a non-string type , take for example a map[string]interface{} from file path(input), as it can't just returned as other type because [input.go#101](https://github.com/AlecAivazis/survey/blob/11174a29b34e682c8045cf066601c6343e8e7c72/input.go#L101) will panic).

```go
import "ithub.com/AlecAivazis/survey/v2"
import "ithub.com/AlecAivazis/survey/v2/core"

// [...]

err := survey.Ask(qs, &v);
if err != nil {
  if name, ok := core.IsFieldNotMatch(err); ok {
    // [...name is the not matched question name]
  }
}
```

Please accept this PR as soon as possible, it 
- **does not change any current behavior**
- **does not extend the public API except `/core`**

Happy Holidays!

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -12,7 +12,7 @@ import (
 // the tag used to denote the name of the question
 const tagName = "survey"
 
-// Settables allow for configuration when assigning answers
+// Settable allow for configuration when assigning answers
 type Settable interface {
 	WriteAnswer(field string, value interface{}) error
 }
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Hey @kataras! Happy holidays to you aswell.

I think this is a fine addition to `survey/core`. Mind adding some tests for the new logic?

kataras:
> Hey @kataras! Happy holidays to you aswell.
> 
> I think this is a fine addition to `survey/core`. Mind adding some tests for the new logic?

There is no change on the logic. Just a typed error instead of a string-based error.

AlecAivazis:
@kataras there is in fact new logic. `func (err errFieldNotMatch) Is(target error)` and the `IsFieldNotMatch` are both new functions and methods. I would like to maintain as good coverage as I can and both of these are very easily testable

AlecAivazis:
@kataras - I'm a bit surprised by your response. I hoped that the responsibility act includes test coverage and actually seeing a PR through

I'm going to merge this but you clearly did not read the contribution guidelines and have little intention of adding tests to cover the new logic. Please understand that i have to maintain your contribution and tests help me do that.

PR #271, opened 2020-02-12
Panic index out of range; selectedIndex should not be a negative number

jcpowermac:
When using AskOne, filtering and using arrow keys can result in
a panic.  Add checks to make sure that selectedIndex does
not contain a negative number.

### DIFF
--- a/select.go
+++ b/select.go
@@ -151,6 +151,12 @@ func (s *Select) OnChange(line []rune, pos int, key rune) (newLine []rune, newPo
 	if len(options) <= s.selectedIndex {
 		return []rune{}, 0, false
 	}
+
+	// s.selectedIndex should not be a negative number
+	if s.selectedIndex < 0 {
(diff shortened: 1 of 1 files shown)


jcpowermac:
Additional details in BZ if curious: https://bugzilla.redhat.com/show_bug.cgi?id=1775569

AlecAivazis:
Would you mind cherry picking this onto v2?

jcpowermac:
@AlecAivazis I have test code that I can try to see if I can reproduce with v2.  There were differences so I don't think it will cleaning CP.

AlecAivazis:
Ah that might definitely be the case. If you have the time to verify, that would save me having to add another item to my list of stuff to do 🙌

jcpowermac:
@AlecAivazis tested with v2 no panic.

AlecAivazis:
awesome! Thanks for checking

PR #276, opened 2020-02-18
multiselect: `enter` to `space` in help

gobwas:
Looks like this must be space.


AlecAivazis:
Thanks for fixing this @gobwas! Looks like you forgot to update the corresponding test to reflect your change.

PR #284, opened 2020-03-27
[WIP] Failing test for default not in options in Select prompt

antonio-osorio:
Hi @AlecAivazis the behavior of the Select prompt when the default value is not one of the options is a bit counter intuitive. I have added a test that shows the failure.

For a select with options "one", "two", default="invalid default" the user sees:

```
Pick one option:
> one
  two
```

And if they hit enter, the resulting value is "invalid default"

What would be the preferred behavior? Thoughs:
- Ignore the default and return item at index 0
- Validate that the default is in the options and return an error if it is not.

### DIFF
--- a/select_test.go
+++ b/select_test.go
@@ -165,6 +165,22 @@ func TestSelectPrompt(t *testing.T) {
 			core.OptionAnswer{Index: 2, Value: "green"},
 		},
 		{
+			"default not in options",
+			&Select{
+				Message: "Choose a color:",
+				Options: []string{"red", "blue", "green"},
+				Default: "",
+			},
+			func(c *expect.Console) {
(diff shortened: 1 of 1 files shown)


szechyjs:
I'm running into this same issue, I would prefer it to select the item at index 0, as that's what simply pressing enter makes it seem like its doing.

AlecAivazis:
I think this makes sense since it matches the user's actions. Nice recommendation and thanks for the test!

AlecAivazis:
@antonio-osorio any update?

AlecAivazis:
Going to close this as its gone a bit stale. if someone wants to pick it up ill gladly review a PR

PR #292, opened 2020-06-25
[#101] Fix survey duplicate lines bug by getting the width of the terminal

tomercy:
This PR solves issue #101 .
Created a new function to determine the terminal width and conclude the number of lines the input takes before calling the "resetPrompt" function.

### DIFF
--- a/confirm_test.go
+++ b/confirm_test.go
@@ -7,8 +7,8 @@ import (
 	"os"
 	"testing"
 
-	"github.com/AlecAivazis/survey/v2/core"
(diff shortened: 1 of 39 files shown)


AlecAivazis:
Hey @tomercy thanks for the submission! It looks like this relies on the `stty` command which unforuntately does not exist on windows machines. Have you tried this on `cmd.exe` or any other windows console?

tomercy:
My solution actually caused another bug so I closed it.... ^^

PR #293, opened 2020-07-09
Write OptionAnswer value to non interface{} map

infixint943:
Hello!
Basically, the commit enables to write just the string value (or index) of an option selected in `survey.Select{}` to a map of `strings`/`ints`. This is achieved by passing a `map[string]string` or `map[string]int` instead of `map[string]interface{}` .

Also, thanks a lot for this wonderful package!

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -86,7 +86,17 @@ func WriteAnswer(t interface{}, name string, v interface{}) (err error) {
 		return copy(field, value)
 	case reflect.Map:
 		mapType := reflect.TypeOf(t).Elem()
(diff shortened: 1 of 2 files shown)


infixint943:
@coryb @AlecAivazis Could you please review this PR, and suggest changes if any?

AlecAivazis:
Hey @infixint943! Would you mind writing a test that covers this change? I am having a hard time following the exact situation you are trying to address.

infixint943:
@AlecAivazis I have added tests.
I would be very thankful if you reviewed them!

infixint943:
Any updates @AlecAivazis

Issue #307, opened 2020-11-05
Selection list not scrolled automatically when survey.WithPageSize is used


pantelis-karamolegkos:
**What operating system and terminal are you using?**

MacOS Catalina / iterm2

**An example that showcases the bug.**

If the list from which the user is supposed to select spans multiple lines and exceeds the screen size, navigating with Up/Down arrows does not re-center the list (the position of the cursor stops being visible if one scrolls to items above/below the screen size).

Here is the code I am running 

```
package main

import (
	. "fmt"

	"github.com/AlecAivazis/survey/v2"
)

func main() {
	var selectedDay string
	myArray := []string{"Monday", "Tuesday", "Wednesday", "Thursday", "Friday", "Saturday", "Sunday"}
	prompt := &survey.Select{
		Message: "Choose a day:",
		Options: myArray,
	}
	survey.AskOne(prompt, &selectedDay, survey.WithPageSize(len(myArray)))
	Println(selectedDay)
}
```

and here is a depiction of the issue...

![Screenshot 2020-11-05 at 10 21 05 PM](https://user-images.githubusercontent.com/50835081/98292319-372a6900-1fb5-11eb-9b20-baf0ac9db8ee.png)

I have deliberately made the `iterm2` terminal hight very small.

When I go above `Thursday` entry,  the entries above (i.e. `Monday`, `Tuesday`, `Wednesday`) do not come up on the screen, they do not become visible. (since the cursor does not re-focus)

I believe this happens only when `survey.WithPageSize(len(myArray))` is used.

**What did you expect to see?**

The list scrolling and me being able to view the options

**What did you see instead?**

The cursor gets lost. Options outside the screen span never become visible.

AlecAivazis:
@Thanks for reporting this @pantelis-karamolegkos! I think this bug was introduced as part of a recent PR so i'm going to revert that now so you can test it for me :)

AlecAivazis:
I just published the revert under `v2.2.1`. Please let me know if you run into more problems :)

pantelis-karamolegkos:
> I just published the revert under `v2.2.1`. Please let me know if you run into more problems :) Ill wait for you to confirm before i close this ticket

Thank you! I will give it a try and report back if I encounter the same issue 👍

pantelis-karamolegkos:
I think I am having the same issue...Can you check the recording? I have my `go.mod` file open to verify I am using the l8st version. Either way you can try and run my code in an iterm2  window small enough for the purpose.

[...]

PR #308, opened 2020-11-05
reverted konsole support which broke runereader

AlecAivazis:
This PR addresses a bug reported in #307 and reverts the changes made by @System-Glitch which seems to have broken runereader in both windows and the default mac terminal.

Unfortunately this means we don't currently support Konsole but that can come after this gets fixed.

cc @pantelis-karamolegkos


System-Glitch:
I will look into it and try to fix it. Thank you for mentioning me.

PR #309, opened 2020-11-06
Fix PreviousLine and NextLine on linux

System-Glitch:
References:

#300 #302 #307 #308 

I think this change will be enough to make it work on both Windows and Linux. I have tested in cmd.exe on windows, and on Ubuntu Konsole and VSCode integrated terminal. **I cannot test on MacOS though because I don't own a Mac**.

**Logic behind this change**:

In `cursor_windows.go`, `PreviousLine()` calls `Down()` and `HorizontalAbsolute()`, so I just did the same for linux but without `HorizontalAbsolute()` because it is apparently not needed (Test it on MacOS and tell me if it works. If it doesn't, then I think adding a call to `HorizontalAbsolute()` would do the trick). This allows us to keep `PreviousLine()` and `NextLine()` and don't have a conflict in naming between the windows and other versions.

### DIFF
--- a/terminal/cursor.go
+++ b/terminal/cursor.go
@@ -42,12 +42,12 @@ func (c *Cursor) Back(n int) {
 
 // NextLine moves cursor to beginning of the line n lines down.
 func (c *Cursor) NextLine(n int) {
-	fmt.Fprintf(c.Out, "\x1b[%dE", n)
+	c.Down(1)
 }
 
 // PreviousLine moves cursor to beginning of the line n lines up.
 func (c *Cursor) PreviousLine(n int) {
-	fmt.Fprintf(c.Out, "\x1b[%dF", n)
+	c.Up(1)
 }
 
 // HorizontalAbsolute moves cursor horizontally to x.


AlecAivazis:
Thanks for getting this out so quickly. I am a bit confused tho, your original PR made the switch from `NextLine` to `Down` - doesn't this change have the same effect as before?

System-Glitch:
@AlecAivazis My previous PR replaced all occurrences of function calls for `PreviousLine()` and `NextLine()` with `Up()` and `Down()`. For Linux terminals, the result is the same. However, an unforeseen effect of that was that is broke the lib for windows because on windows you need `HorizontalAbsolute()`, which was not called anymore because of the replacement.

AlecAivazis:
@System-Glitch - finally got my hands on a mac laptop. I will pull this down now and give it a shot

System-Glitch:
@AlecAivazis I got a friend of mine test it for me on mac and everything appears to be working correctly.

AlecAivazis:
I was also able to confirm that the behavior is consistent across mac and windows. Thanks for being patient 👍

Issue #305, opened 2020-10-08
Put cursor on selected line


shindere:
This issue stems from a discussion started at

https://github.com/cli/cli/issues/1085

and in particular comment

https://github.com/cli/cli/issues/1085#issuecomment-705164274

Would it please be possible to patch the library so that the cursor gets
placed on the selected line? It would make it more usable from a braille
display which generally follows the cursor.

AlecAivazis:
Absolutely! Thank you so much for bringing this up. I had NO idea that's all it took to work better with braille displays.

shindere:
Alec Aivazis (2020/11/02 13:31 -0800):
> Absolutely! Thank you so much for bringing this up. I had NO idea
> that's all it took to work better with braille displays.

Many thanks for your positive feedback!

This kind of knowledge is indeed not very widespread.

If you implement the feature and need testing, please feel free to
contact me either here or by e-mail directly:
Sebastien.Hinderer@inria.fr

Thanks!

vilmibm:
@AlecAivazis Hiya!

I'm hoping to take on this work on behalf of the GitHub CLI team; let me know if there are pitfalls or other issues I should be aware of before I dig it.

AlecAivazis:
Awesome!! I’m super excited to hear that @vilmibm. Nothing specific comes to mind but please let me know if you need any guidance.

PR #312, opened 2020-11-16
Prevent cursor from being hidden

apolitis-work:
The cursor was hidden during input. This allows the cursor to be displayed.

### DIFF
--- a/input.go
+++ b/input.go
@@ -149,8 +149,7 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
 	defer rr.RestoreTermMode()
 
 	cursor := i.NewCursor()
-	cursor.Hide()       // hide the cursor
-	defer cursor.Show() // show the cursor when we're done
+	cursor.Show() // show the cursor when we're done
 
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Hey @apolitis-work!

Mind giving me a little more context about this change? at the moment the cursor doesn't really behave as it "should" so i think i'm inclined to leave it hidden until we address #305

apolitis-work:
Sure thing! In our CLI application, we just wanted to have the cursor show up on the line which we were typing our input on. This fix seemed to make that happen, and I figured I would just submit a PR to see if you thought it was worth merging.

So ideally after #305 a user could just press up and down to move the cursor onto a selected line? I suppose my change would break that. I'm fine with leaving the cursor hidden if I'm interpreting that correctly

AlecAivazis:
Yea that's the intent. Screenreaders will read the line that the cursor is on so we really do need to move the cursor so that it always lines up correctly. I'll go ahead and close this PR to keep things tidy. thanks for opening it!

PR #316, opened 2020-11-26
Error handling for reading runes on multiselect

soprea:
SIGHUP is ignored and the program gets into a weird loop waiting for select to happen.
Steps to reproduce:
`go run examples/longmulti.go`
In a different terminat run `kill -s hup (pidof longmulti)` Or simply close the terminal while the program waits for input.
The program is not killed and gets in a weird state slowly killing your CPU

### DIFF
--- a/multiselect.go
+++ b/multiselect.go
@@ -273,7 +273,10 @@ func (m *MultiSelect) Prompt(config *PromptConfig) (interface{}, error) {
 
 	// start waiting for input
 	for {
-		r, _, _ := rr.ReadRune()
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Oh damn, nice catch! Do you mind looking at the other prompts to make sure we dont have the same problem?

soprea:
Already checked and also tested. They all look good. Only multiselect got missed.

AlecAivazis:
Sweet! Thanks for fixing this

Issue #298, opened 2020-08-26
Modify prompt text color


droctothorpe:
Is there a way to modify the Select question prompt text color?

I see https://github.com/AlecAivazis/survey/blob/9f0147aa9ef6a6810ebb976da58f2c331a96b5a9/select.go#L48 but I'm not entirely sure what to do with it. An example would be really helpful.

Thank you! 

PS. This library is amazing and I sing it's praises to all the Golang devs I know who are willing to listen. Thank you for all of the hard work that you put into it.

droctothorpe:
NVM! Figured this out. Might be worth changing this heading:

https://github.com/AlecAivazis/survey/blob/863da4d0600f0b2e969352c91a808e5bc94e220d/README.md#changing-the-icons

to `Change the Icons and Prompt Text Color` or something to that effect.

droctothorpe:
Hm not seeing a way to modify the default help text `Use arrows to move, type to filter` color though:
![image](https://user-images.githubusercontent.com/24783969/91325957-c6372b80-e791-11ea-85b4-0c7f5190a99a.png)

That could be a good thing to make configurable.

AlecAivazis:
Oh! I'm not sure how I missed this issue. I'm sorry for not responding sooner and thank you for the kind words!

I think supporting that level of modifcation is probably too-fine grained for my personal use but I dont really mind adding yet another field to an IconSet for the default color.  "`IconSet`" doesn't really fit anymore, but oh well it's not worth breaking the API for something so small

kenny01123:
Hi! I just made the PR according to this issue. Now you can change the help text color like this in select:

`color := ""
prompt := &survey.Select{
    Message: "Choose a color:",
    Options: []string{"red", "blue", "green"},
    HelpTextColor: "red",

}
survey.AskOne(prompt, &color)`

PR #317, opened 2020-11-27
Help Text Color configurable for users

kenny01123:
Hi My Name is Kenny and I want to help with this project.
I added some code to make the help text color configurable with respect to #298 
the UX with the added functionality should be the same, 
as you can still have the default cyan color help text without any changes.

For now its only in select.go. If i get a nod i can also expand it to other prompts.

Thank you!

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,6 +414,7 @@ The icons and their default text and format are summarized below:
 | SelectFocus    | >    | green      | Marks the current focus in `Select` and `MultiSelect` prompts |
 | UnmarkedOption | [ ]  | default+hb | Marks an unselected option in a `MultiSelect` prompt          |
 | MarkedOption   | [x]  | cyan+b     | Marks a chosen selection in a `MultiSelect` prompt            |
+| PromptText     | []   | cyan       | Marks a chosen selection in a `MultiSelect` prompt            |
 
 ## Custom Types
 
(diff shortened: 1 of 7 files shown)


AlecAivazis:
hey @kenny01123! Thanks for your contribution.

 I think this kind of configuration should be passed as a `survey.Option` to `Ask`/`AskOne`, similar to what we do for [Changing the Icons used by a prompt](https://github.com/AlecAivazis/survey#changing-the-icons)

kenny01123:
Hey @AlecAivazis I added a field called Prompt Text in IconSet.
This is added to `select`, and i have changed the help text color configuration in `multiselect` and `multiline` from 
`{{- color $.Config.Icons.SelectFocus.Format}}` to `{{- color $.Config.Icons.PromptText.Format}}`. 
The reason is that the `select focus` imo should only be referred as the selected option in the list but not the helptext.

I only set the format as the option in the field but not the text. Since `select`, ``multiselect` and `multiselect` have different default texts.

What is your opinion about it? 
#317

AlecAivazis:
Hey @kenny01123!

I think these changes make sense. It could be a little confusing specifying the prompt text as an `Icon` since the `Text` field is ignored but i'm not sure its worth introducing anything new and it is related to the rest of the `IconSet`. 

Looking forward to seeing the other prompts updated aswell! Thanks again for the contribution 👍

AlecAivazis:
I'm going to close this PR to keep things tidy, please feel free to open another one up when you are ready

PR #321, opened 2020-12-17
Fix re-rendering when prompt is exactly the width of the terminal

### LINKED ISSUE #318: Typing into prompt clears previous lines in terminal
MacOS Catalina / iTerm. Saw this on MacOS High Sierra as well.

I've attached a video of the bug in this zip file: [BugDemo.zip](https://github.com/AlecAivazis/survey/files/5619166/BugDemo.zip)

It seems that when I adjust the width of my terminal tab so that the prompt text is all the way at the end of the terminal screen (i.e. no whitespace between the last character in the prompt description text - ":" in this case -  and the end of the terminal screen), a line in my terminal is cleared every time I type a character.

I would expect my terminal to remain as is when I type input into a prompt.

**What did you see instead?**
I saw my terminal clear when I typed given the specifications above.

mislav:
The code that measures width of rendered lines of content and whether any have overflown in the terminal did not account for the possibility that the content could be _exactly_ the width of the terminal. In that case, it shouldn't be counted as overflow.

Fixes #318
Followup to #291, #288

### DIFF
--- a/renderer_posix_test.go
+++ b/renderer_posix_test.go
@@ -1,0 +1,81 @@
+// +build !windows
+
+package survey
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	expect "github.com/Netflix/go-expect"
+	"github.com/hinshun/vt10x"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRenderer_countLines(t *testing.T) {
+	t.Parallel()
+
+	termWidth := 72
+	stdout := new(bytes.Buffer)
+	c, _, err := vt10x.NewVT10XConsole(expect.WithStdout(stdout))
+	vt10x.ResizePty(c.Tty(), termWidth, 30)
+	require.Nil(t, err)
+	defer c.Close()
+	defer c.Tty().Close()
+
+	r := Renderer{stdio: Stdio(c)}
+
+	tests := []struct {
+		name  string
+		buf   *bytes.Buffer
+		wants int
+	}{
+		{
+			name:  "empty",
+			buf:   new(bytes.Buffer),
+			wants: 0,
+		},
+		{
+			name:  "no newline",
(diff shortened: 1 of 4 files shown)


renderer_posix_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestRenderer_countLines(t *testing.T) {
+	t.Parallel()
+
+	termWidth := 72
+	stdout := new(bytes.Buffer)
+	c, _, err := vt10x.NewVT10XConsole(expect.WithStdout(stdout))
+	vt10x.ResizePty(c.Tty(), termWidth, 30)
```
mislav, line 22: This test is marked as POSIX-only because it allocates a pseudo-terminal to emulate a scenario when the terminal is of a specific width. If anyone knows of a more lightweight approach to generate a terminal at a file descriptor that pretends to be a certain size, I'm all ears!

AlecAivazis:
Thanks for fixing this @mislav - i verified the behavior on both apple and windows.

Thanks for being patient while we wrangle this! Maintaining this project alone takes a lot of time and attention, two things i haven't had much of in the past month.

AlecAivazis:
On that note, are there any extra resources at GitHub that can help me test PRs and releases? There have been two regressions in the past month that are mostly due to me juggling multiple things at once, i would like to find a way to make sure that doesn't happen again

mislav:
@AlecAivazis Sure, let us know how we can help. We are just 3 devs and don't have infinite resources either, but we rely quite a lot on Survey and we'd like to have it be stable. A lot of problems that our users reported with GitHub CLI were fixed over the last few releases of Survey, and we appreciate that!

We are also wondering whether you would be open to contributions that improve accessibility by positioning the cursor at the right place, and possibly allowing custom keybindings to Prompt components like the Editor.

mislav:
BTW, thank you for making a patch release in such a short time! 🙇 ❤️

Issue #320, opened 2020-12-17
Flakey test: TestOnInterruptFunc


mislav:
**What operating system and terminal are you using?** any non-Windows platform

**An example that showcases the bug.**
```sh
$ while true; do go test -count=1 -run ^TestOnInterruptFunc$ github.com/AlecAivazis/survey/v2 || break; done
```

**What did you expect to see?** test always passes

**What did you see instead?**
```
ok  	github.com/AlecAivazis/survey/v2	0.085s
ok  	github.com/AlecAivazis/survey/v2	0.085s
ok  	github.com/AlecAivazis/survey/v2	0.086s
ok  	github.com/AlecAivazis/survey/v2	0.102s
ok  	github.com/AlecAivazis/survey/v2	0.240s
Ended abruptly!
The end.
--- FAIL: TestOnInterruptFunc (0.00s)
    --- FAIL: TestOnInterruptFunc/Global_OnInterrupt (0.00s)
        survey_test.go:350: Raw output: "? Are you a bot?^C\r\n\x1b[?25l\x1b[?25h? Are you a bot? \r\n"
        survey_test.go:350:
            ? Are you a bot?^C
            ? Are you a bot?
        Error Trace:    survey_test.go:360
        Error:      	Not equal:
                    	expected: *errors.errorString(&errors.errorString{s:"interrupt"})
                    	actual  : <nil>(<nil>)
        Test:       	TestOnInterruptFunc/Global_OnInterrupt
FAIL
FAIL	github.com/AlecAivazis/survey/v2	0.463s
FAIL
```

This is caused by a race condition in a test added in #301 that spawns several `t.Run()` blocks, which themselves run in separate goroutines that enable parallel test runs by invoking the `RunTest()` helper, but these tests also depend on the state of the global package-level variable OnInterrupt that changes value during the test.

I could easily fix this test in a PR, but I would argue that https://github.com/AlecAivazis/survey/pull/301 should be [reverted instead](https://github.com/AlecAivazis/survey/issues/319). /cc @infinitepr0

infinitepr0:
You're saying a global level `OnInterrupt` is a bad idea?

infinitepr0:
I don't see any need to revert the PR, since there's no breaking changes. The panic that happened in #319 is because I failed to account for the case when OnInterrupt function is not set, which is fixed by #322. I still don't get how it managed to not panic on test "No OnInterrupt" tho.

Maybe you could help me understand why it didn't FAIL?

[...]

PR #322, opened 2020-12-17
Fix regression in PR #301.

### LINKED ISSUE #319: New OnInterrupt behavior causes crash on Ctrl-C with default settings
**What operating system and terminal are you using?** irrelevant

**An example that showcases the bug.** `go run examples/inputfilesuggestion.go` & press <kbd>Ctrl-C</kbd>

**What did you expect to see?** "interrupt" being printed and program exits

**What did you see instead?** crash on this line: https://github.com/AlecAivazis/survey/blob/f45c05b5039ca1f9c9c562be232f1c4044e0add0/survey.go#L311-L312

This is a regression in v2.2.5 due to the new OnInterrupt feature #301.

@AlecAivazis Why did a new, poorly argued for, backwards-incompatible feature ship in a patch release? Can we expect more functional changes like this in the v2.2.x line?

Would you consider branching off something like a `release-2.2` branch, reverting #301 there, and backporting to it only bug fixes going forward so that we can be sure that new tags on the v2.2.x patch line do _not_ also bundle new features?

infinitepr0:
Fix #319

### DIFF
--- a/survey.go
+++ b/survey.go
@@ -307,8 +307,8 @@ func Ask(qs []*Question, response interface{}, opts ...AskOpt) error {
 
 		// grab the user input and save it
 		ans, err := q.Prompt.Prompt(&options.PromptConfig)
-		// if SIGINT is recieved.
-		if err == terminal.InterruptErr {
+		// if SIGINT is recieved and OnInterrupt function provided.
+		if err == terminal.InterruptErr && options.OnInterrupt != nil {
 			options.OnInterrupt()
 		}
 		// if there was a problem


infinitepr0:
Regarding #320, I feel running `t.Parallel()` should be done in each test function, rather than the `RunTest` function. However, if there is a strong reason to NOT do the same, I can fix the issue by wrapping each test in `TestOnInterruptFunc` inside another `t.Run()`.

infinitepr0:
Also, isn't the minor version supposed to be incremented, rather than the patch (I believe you're using semver) ?

AlecAivazis:
I reverted the other PR and published onto the `v2.2.x` branch in order to correct version issue. So at the very least we will need to bring the original changes into this PR. 

That being said, I'm not sure the change makes a lot of sense. Thinking about what @mislav said, i agree that this change is pretty unnecessary given that the caller can just look to see if the returned error is an interrupt? Especially considering that the error is returned either way, i can't really see the weight of an addition to the top-level API being worth it.

I'm inclined to close this PR but I want to give you a change to explain if I am missing something.

infinitepr0:
See https://github.com/AlecAivazis/survey/pull/323#issuecomment-748615810

Also, if top level API's are a problem, I could remove the top level API and leave only the `Question` stuct level API.

infinitepr0:
This PR is being closed in correspondence to https://github.com/AlecAivazis/survey/pull/323#issuecomment-749516182.

PR #327, opened 2021-01-10
Use rune when delete or filter answer

AielloChan:
When using Unicode in the `string` type answer will cause a problem.

Example:

```bash
> go run examples/simple.go
? What is your name? 小明
```

### DIFF
--- a/input.go
+++ b/input.go
@@ -105,7 +105,8 @@ func (i *Input) OnChange(key rune, config *PromptConfig) (bool, error) {
 		}
 	} else if key == terminal.KeyDelete || key == terminal.KeyBackspace {
 		if i.answer != "" {
-			i.answer = i.answer[0 : len(i.answer)-1]
+			runeAnswer := []rune(i.answer)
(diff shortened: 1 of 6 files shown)


AlecAivazis:
Hey @AielloChan! Thanks for submitting this fix, would you mind adding a test that shows the problem is fixed?

AielloChan:
Check if that is ok😬

AlecAivazis:
Okay, I finally found some time to pull this down and give it a shot. Thanks for being patient!

Everything seems to behave correctly on my machine, however since this touches a bunch of prompts, i'd like to get a second pair of eyes on it. Maybe @coryb or @mislav have some spare time?

AlecAivazis:
Hm, I might be missing something but wouldn't this also affect the `Select` prompt?

AielloChan:
> Hm, I might be missing something but wouldn't this also affect the `Select` prompt?

Yes, because prompt filter also need use `Rune` for unicode words.

mislav (approved):
This looks good to me!

AlecAivazis:
great! thanks for taking a look @mislav

Issue #329, opened 2021-01-16
[Bug] ComposeTransformers fails on optional input (zero value)


Roytangrb:
**What operating system and terminal are you using?**
OS: `macOS Catalina Version 10.15.6`
Terminal: `zsh`
go version: `go1.15.6 darwin/amd64` 

**An example that showcases the bug.**
A survey question asking for an optional string, if `survey.ComposeTransformers` is used, and user skip the question, the program panics:
```go
{
	Name: "scope",
	Prompt: &survey.Input{
		Message: "Enter scope:", // an optional string input
	},
	Transform: survey.ComposeTransformers(
		survey.TransformString(strings.ToLower),
		survey.TransformString(strings.TrimSpace),
	),
},
```
Or to reproduce more straight forward:
```go
t1 := survey.TransformString(strings.ToLower)
t2 := survey.TransformString(strings.TrimSpace)

var s string
fmt.Println(t1(s)) //<nil>
fmt.Println(t2(t1(s))) // panic
```

**What did you expect to see?**
Transformers should be skipped if value is zero value

**What did you see instead?**
```
panic: reflect: call of reflect.Value.Interface on zero Value

goroutine 1 [running]:
reflect.valueInterface(0x0, 0x0, 0x0, 0xc000051801, 0x13b0f01, 0xc000051820)
	/usr/local/opt/go/libexec/src/reflect/value.go:1021 +0x1ad
reflect.Value.Interface(...)
	/usr/local/opt/go/libexec/src/reflect/value.go:1016
github.com/AlecAivazis/survey/v2.isZero(0x0, 0x0, 0x0, 0x1)
	/Users/roytang/.go/pkg/mod/github.com/!alec!aivazis/survey/v2@v2.2.7/validate.go:86 +0x8e
github.com/AlecAivazis/survey/v2.TransformString.func1(0x0, 0x0, 0x0, 0x0)
	/Users/roytang/.go/pkg/mod/github.com/!alec!aivazis/survey/v2@v2.2.7/transform.go:20 +0xaf
```

Roytangrb:
Hi @AlecAivazis, I reckon the issue is caused by `nil` return by `TranformString` zero value check. The `nil` is passed to the next Transformer while the Composed Transformer execute. Please review my PR that might fix, thanks!

PR #330, opened 2021-01-16
fix(transform): fix nil returned by TransformString

Roytangrb:
TranformString should not return `nil` on zero value, otherwise `nil` will be passed to next Tranformer if there are more than 1 Transformer in ComposeTransformers.

In next TranformString, while checking if if value is zero at https://github.com/AlecAivazis/survey/blob/e64d9b644fb7d85d69cd3e33a92a43ee706c1b6f/validate.go#L86

`reflect.ValueOf(nil).Interface()` and `reflect.Zero(reflect.ValueOf(nil).Type()).Interface()`will cause program to panic

To fix:
return zero value of string (`""`) instead coz String Tranformer is guaranteed to return string

fix https://github.com/AlecAivazis/survey/issues/329

### DIFF
--- a/transform.go
+++ b/transform.go
@@ -18,18 +18,21 @@ func TransformString(f func(s string) string) Transformer {
 	return func(ans interface{}) interface{} {
 		// if the answer value passed in is the zero value of the appropriate type
 		if isZero(reflect.ValueOf(ans)) {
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Nice catch! Thanks for submitting a fix

PR #335, opened 2021-02-12
Fixes #334: Make sure the cursor goes back to the begining of the lin…

### LINKED ISSUE #334: Cursor is sometimes not going back to the start of the line

macOS Big Sur 11.0.1 with iterm2

```
package main

import (
	"fmt"
)

func main() {
	prompt := &survey.Password{
		Message: "password",
	}
	var pass string
	_ = survey.AskOne(prompt, &pass)

	fmt.Print("hello")
}
```

When running this sample, one can see that after the prompt the cursor stays aligned with the previous prompt text:

Screenshot 2021-02-12 at 16 10 10

The cursor is going back to the beginning of the line.

I tracked the issue to PR #309. I re-ran the previous test by adding a call to `c.HorizontalAbsolute(0)` (like it's done for Windows) and confirmed this was fixing my issue.

PierreBtz:
…e when jumping lines.

See #334 for the details. This code is doing the same as what Windows does.

There is something odd in the the n parameter passed to both methods is never used as we use 1. The Windows code does not do this and the previous code (changed in #309) did not do this either. I have a feeling it should be `c.Down(n)` and `c.Up(n)` but I did not analyze this library enough to be confident.

### DIFF
--- a/terminal/cursor.go
+++ b/terminal/cursor.go
@@ -43,11 +43,13 @@ func (c *Cursor) Back(n int) {
 // NextLine moves cursor to beginning of the line n lines down.
 func (c *Cursor) NextLine(n int) {
 	c.Down(1)
+	c.HorizontalAbsolute(0)
 }
 
 // PreviousLine moves cursor to beginning of the line n lines up.
 func (c *Cursor) PreviousLine(n int) {
 	c.Up(1)
+	c.HorizontalAbsolute(0)
 }
 
 // HorizontalAbsolute moves cursor horizontally to x.


AlecAivazis:
@coryb @mislav any chance you could test this out? I'm always nervous being the only set of eyes on something touching cursor interactions

nhatthm:
Thanks @PierreBtz for the fix 🙌

I could reproduce the issue on my Mac and with [hinshun/vt10x](https://github.com/hinshun/vt10x) (see [my test](https://github.com/nhatthm/surveydog/pull/1)). It's fixed after applying your changes 👍

However, I don't have experience in Cursor and the test was in the emulator so I can't validate your code :D

AlecAivazis:
@PierreBtz - i fixed the issue preventing the CI pipeline from executing. Mind rebasing onto master so we can get this in?

AlecAivazis:
And thanks @nhatthm for verifying this fix!

PierreBtz:
@AlecAivazis I rebased, looks like the build is passing now, thanks for fixing it, and thanks for your work on this project :)

AlecAivazis (approved):

PR #351, opened 2021-04-16
fix(input.go): delete answer after validate failure

### LINKED ISSUE #350: after validate failure, answer value not output or clean
macos iterm2 zsh

```golang
package main

import (
	"errors"
	"fmt"
	"net/url"

)


func main() {
	prompt := &survey.Input{Message: "please input a url"}
	var answer url.URL
	err := survey.AskOne(prompt, &answer,survey.WithValidator(urlValidate))
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Println(answer)
}

func urlValidate(val interface{}) error {
	urlStr, ok := val.(string)
	if !ok || urlStr == "" {
		return errors.New("please input correct url")
	}
	res, err := url.Parse(urlStr)
	if err != nil || res.Scheme == "" || res.Host == "" || res.Path == "" {
		return errors.New("please input correct url")
	}
	return nil
}

```

Untitled
i inputed "aaa", then press Enter, the validate not passed, and the answer output  has cleard, but when i continue to type "bbb", the "aaa" will show, continue press Enter, The answer is once again cleared, but when i input "http", the "aaabbb" will show. I think this is a confused behavior, is my code wrong? i also tested `survey.Required`, it has same behavior.
**What did you see instead?**
after validate failure, old answer should show or clean

zigitn:
Closes #350

### DIFF
--- a/input.go
+++ b/input.go
@@ -187,7 +187,7 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
 	lineStr := i.answer
 
 	i.AppendRenderedText(lineStr)
-
+	i.answer = ""
 	// we're done
 	return lineStr, err
 }


AlecAivazis:
@zigitn - i fixed the issue preventing the CI pipeline from executing. Mind rebasing onto master so we can get this in?

zigitn:
Of course, do I need to submit a new pr?

AlecAivazis:
You should be able to get away with just force-pushing over your existing branch

AlecAivazis:
Thanks for doing that @zigitn! Do you mind adding a test that veriies this behavior?

cben:
@AlecAivazis I didn't find existing tests for `WithValidator` that actually test the input.go logic.  Do you have advice how/where to test this?

AlecAivazis:
@cben - thanks for picking this up. I think the [`Ask` tests](https://github.com/AlecAivazis/survey/blob/c79c4a988cde2621cf87a59265233ed7e9e0b63e/survey_test.go#L309) are a good example to copy

AlecAivazis:
Going to close this as its gone a bit stale

PR #354, opened 2021-05-11
Re-enable TestAsk

cben:
Those got disabled in https://github.com/AlecAivazis/survey/pull/216/files#diff-3a029c8ea50304c9dc5637560df2375bc28f1620d72744a13f4d3c152bb16a4aR120 and apparently were forgotten since then?

I noticed several tests, some here and some in other files, _sometimes_ get stuck when running `go test -v`.  
Everything works reliably without `-v`, so I guess there are timing issues due to verbose printing; but when focusing on a single test with `-run=REGEXP` they also work pretty reliably even with `-v`.

- Exception: I kept the "What days do you prefer" test commented out because it always gets stuck, with or without `-v`, and I don't know how to debug it (`-v` doesn't help — it doesn't get to show "raw output" which is usually very helpful).

### DIFF
--- a/survey_test.go
+++ b/survey_test.go
@@ -150,12 +150,14 @@ func TestAsk(t *testing.T) {
 				{
 					Name: "commit-message",
 					Prompt: &Editor{
+						Editor:  "vi",
 						Message: "Edit git commit message",
 					},
 				},
 				{
 					Name: "commit-message-validated",
 					Prompt: &Editor{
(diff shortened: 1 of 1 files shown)


cben:
@AlecAivazis please review.

AlecAivazis:
Thanks for hunting this down @cben!

AlecAivazis (approved):

PR #355, opened 2021-05-26
Use `golang.org/x/term` instead of `golang.org/x/crypto/ssh/terminal`

tklauser:
The `golang.org/x/crypto/ssh/terminal` package is deprecated and merely a
wrapper around `golang.org/x/term`. Use the latter directly.


AlecAivazis:
Thanks for updating this @tklauser!

PR #358, opened 2021-06-16
first pass on cursor tracking select focus

### LINKED ISSUE #305: Put cursor on selected line

vilmibm:
Fixes #305

This solution is inelegant and brittle but it does work :sweat_smile: please let me know if I should add more test coverage.

### DIFF
--- a/multiselect.go
+++ b/multiselect.go
@@ -45,9 +45,27 @@ type MultiSelectTemplateData struct {
 	ShowHelp      bool
 	PageEntries   []core.OptionAnswer
 	Config        *PromptConfig
+
+	// These fields are used when rendering an individual option
+	CurrentOpt   core.OptionAnswer
+	CurrentIndex int
+}
+
+// IterateOption sets CurrentOpt and CurrentIndex appropriately so a multiselect option can be rendered individually
+func (m MultiSelectTemplateData) IterateOption(ix int, opt core.OptionAnswer) interface{} {
+	copy := m
+	copy.CurrentIndex = ix
+	copy.CurrentOpt = opt
+	return copy
 }
 
 var MultiSelectQuestionTemplate = `
+{{- define "option"}}
+    {{- if eq .SelectedIndex .CurrentIndex }}{{color .Config.Icons.SelectFocus.Format }}{{ .Config.Icons.SelectFocus.Text }}{{color "reset"}}{{else}} {{end}}
+    {{- if index .Checked .CurrentOpt.Index }}{{color .Config.Icons.MarkedOption.Format }} {{ .Config.Icons.MarkedOption.Text }} {{else}}{{color .Config.Icons.UnmarkedOption.Format }} {{ .Config.Icons.UnmarkedOption.Text }} {{end}}
+    {{- color "reset"}}
+    {{- " "}}{{- .CurrentOpt.Value}}
+{{end}}
 {{- if .ShowHelp }}{{- color .Config.Icons.Help.Format }}{{ .Config.Icons.Help.Text }} {{ .Help }}{{color "reset"}}{{"\n"}}{{end}}
 {{- color .Config.Icons.Question.Format }}{{ .Config.Icons.Question.Text }} {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }}{{ .FilterMessage }}{{color "reset"}}
@@ -56,10 +74,7 @@ var MultiSelectQuestionTemplate = `
 	{{- "  "}}{{- color "cyan"}}[Use arrows to move, space to select, <right> to all, <left> to none, type to filter{{- if and .Help (not .ShowHelp)}}, {{ .Config.HelpInput }} for more help{{end}}]{{color "reset"}}
   {{- "\n"}}
   {{- range $ix, $option := .PageEntries}}
-    {{- if eq $ix $.SelectedIndex }}{{color $.Config.Icons.SelectFocus.Format }}{{ $.Config.Icons.SelectFocus.Text }}{{color "reset"}}{{else}} {{end}}
-    {{- if index $.Checked $option.Index }}{{color $.Config.Icons.MarkedOption.Format }} {{ $.Config.Icons.MarkedOption.Text }} {{else}}{{color $.Config.Icons.UnmarkedOption.Format }} {{ $.Config.Icons.UnmarkedOption.Text }} {{end}}
-    {{- color "reset"}}
-    {{- " "}}{{$option.Value}}{{"\n"}}
+    {{- template "option" $.IterateOption $ix $option}}
   {{- end}}
 {{- end}}`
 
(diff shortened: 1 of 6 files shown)


AlecAivazis:
Hey @vilmibm - thanks for submitting this! I'll try to find some time this weekend to pull it down and give it a shot.

It's been awhile since I've looked at the code for `paginate` but I _think_ you should be able to use the second returned value to figure out how many lines to move up (assuming that the cursor is always starting at the bottom). If that doesn't work for some reason, then you would need to have some other way of identifying the correct line and your approach would make sense. [Here](https://github.com/AlecAivazis/survey/blob/60ad3ef78c56c11e7ede294ddd0fa908b10ff475/survey_test.go#L61) are the tests for paginate so you can get a feeling for how it behaves.

vilmibm:
Using the paginate output is a great idea; I don't know why my eyes just glossed right past the output of `paginate`.

I'll add support for wide option lines and get this cleaned up.

AlecAivazis:
Ah yea, good point. Maybe a reasonable solution is to add a method to `Renderer` that invokes and returns the values from `paginate` along with moving the cursor? That will keep the actual page logic pure and easily testable but still give us the behavior without having to implement it in multiple places. Thoughts?

vilmibm:
Let me know what you think of the current approach; it's not the most elegant (the `width()` method for the prompts is a brittle thing) but I think it keeps things testable / comprehensible.

Unfortunately I broke multi-select rendering so I'll fix that tomorrow; if you are ok with this approach though in general I'll go ahead and add tests as needed.

multiselect.go
```diff
@@ -171,6 +174,18 @@ func (m *MultiSelect) OnChange(key rune, config *PromptConfig) {
...
+	m.OffsetCursor(offset)
+}
+
+func (m *MultiSelect) width() int {
+	w, err := m.termWidth()
+	if err != nil || w == 0 {
+		// assume terminal of arbitrary width
+		w = 10000
+	}
+	return w - 7 // account for leading characters in template
```
AlecAivazis, line 188: Do you think we could look at the template string for this value?

vilmibm: like, the raw template string? I'm staring at it but I'm not sure how to compute what we'd need.

I'm also realizing that the number is wrong if a user configures, say, the select focus icon to ">>" instead of ">" so I think we definitely need to take a different approach there...

I'll think about it tomorrow.

vilmibm: What if we split out the templates? If we have individual line templates for rendering options, we can render each one to determine its rendered length without having to render the entire template and then re-parse it.

AlecAivazis: Oh that's not a bad idea. I'm not against it from a design perspective if it gives us improved accessibility - I think basically _anything_ is worth this kind of improvement.

vilmibm: :+1: thanks for the encouragement. I'll continue with that approach; I got a proof of concept working and am trying to keep the refactoring to an absolute minimum.

vilmibm:
I was waylaid by work stuff but finally marking this as non-draft. This solution is not the best but it does work; it will be a little annoying to support it in new prompt types, but it will weather icons of variable length when computing the width/word wrapping of long options.

AlecAivazis:
Sorry i haven't gotten to this yet @vilmibm. work was really busy this week, will try to find some time to look at it over the weekend

multiselect.go
```diff
@@ -159,18 +174,20 @@ func (m *MultiSelect) OnChange(key rune, config *PromptConfig) {
...
-			Checked:       m.checked,
-			ShowHelp:      m.showingHelp,
-			PageEntries:   opts,
-			Config:        config,
-		},
-	)
+	m.Render(MultiSelectQuestionTemplate, tmplData)
+
+	offset := computeCursorOffset(MultiSelectQuestionTemplate, tmplData, opts, idx, m.termWidthSafe())
+	m.OffsetCursor(offset)
```
AlecAivazis, line 190: What do you think about adding a new method to `Renderer` called somethign like `RenderWithCursorOffset` or something that does these three lines together? It will help dry up the code base as well as give us a single place where accessibility is implemented in case we need to change things down the line.

select.go
```diff
@@ -152,22 +167,33 @@ func (s *Select) OnChange(key rune, config *PromptConfig) bool {
...
+	s.Render(SelectQuestionTemplate, tmplData)
+
+	offset := computeCursorOffset(SelectQuestionTemplate, tmplData, opts, idx, s.termWidthSafe())
+	s.OffsetCursor(offset)
 
 	// keep prompting
 	return false
 }
 
+func (s *Select) width() int {
```
AlecAivazis, line 188: This looks almost exactly like `Renderer.width` - any way we could add an argument to the function that passes the number of characters?

AlecAivazis (changes requested):
Sorry this took me so long to look at! I'm buying a home and have been very busy running back and forth to wrangle all of the moving parts. Thanks for being patient

select.go
vilmibm: oops, that's vestigial; I'll just delete it.

vilmibm:
@AlecAivazis buying a house seems so stressful -- no rush on my part and good luck!

I followed up on those two comments, but I think I have some test fixing to do 😅 I guess because I'm injecting more control sequences the virtual terminal tests need to be reconciled? I'll start work on that; let me know if you have any protips.

AlecAivazis:
Oh damn sorry I didn't realize they were still pending. GitHub didn't notify me about your edit so I totally missed that. I just kicked them off

AlecAivazis (changes requested):
Finally found some time to pull this down and test it. The select and multi select tests all ran smoothly (located at `tests/{select.go, multiselect.go}` but if I ran another test afterwards, the prompt would print many lines above the starting line.  I think there is a problem in the way the select and multiselect prompts clean up

vilmibm:
hm, can you help me with some more context?

I'm running this locally:

` go test -v github.com/AlecAivazis/survey/v2 github.com/AlecAivazis/survey/v2/core github.com/AlecAivazis/survey/v2/terminal -mod=vendor` which is failing for me locally against both `master` and my PR for reasons seemingly unrelated to my changes. It seems like they are passing in CI, though?

How are you triggering the failure?

AlecAivazis:
Ah yea sure. So there is [this directory](https://github.com/AlecAivazis/survey/tree/master/tests) which contains scripts, not actual tests, that are meant to be run manually in order to verify that things behave in a real scenario

vilmibm:
@AlecAivazis ah oops, I screwed up the tiny refactor and was restoring the cursor when I shouldn't have been by accident. Fixed now; running through the stuff in `tests/` seems to work fine.

AlecAivazis:
Awesome! Thanks for hunting that down! I **finally** found the time to pull down your branch and test it out. Sorry that took so long - was on vacation for 10 days.

I'm pretty sure this is good to go but I want to give it one more go around. Pretty excited we should be getting this in soon

AlecAivazis (approved):
I'm going to finally merge this in. Thanks so much for implementing this @vilmibm!

vilmibm:
awesome news! thank you!!

billygriffin:
Great news @AlecAivazis, thanks for working with us to get it across the finish line. And great work @vilmibm! 🎉 ❤️

PR #361, opened 2021-06-28
(fix): problems with input navegation

lucassabreu:
- first pass at input using RuneReader.ReadLine with suggestions
- fix newline from readline
- better callback name

### DIFF
--- a/input.go
+++ b/input.go
@@ -22,6 +22,7 @@ type Input struct {
 	Help          string
 	Suggest       func(toComplete string) []string
 	answer        string
+	typedAnswer   string
 	options       []core.OptionAnswer
 	selectedIndex int
 	showingHelp   bool
@@ -65,6 +66,9 @@ func (i *Input) onRune(config *PromptConfig) terminal.OnRuneFn {
 	return terminal.OnRuneFn(func(key rune, line []rune) ([]rune, bool, error) {
 		if i.options != nil && (key == terminal.KeyEnter || key == '\n') {
 			return []rune(i.answer), true, nil
+		} else if i.options != nil && key == terminal.KeyEscape {
+			i.answer = i.typedAnswer
+			i.options = nil
 		} else if key == terminal.KeyArrowUp && len(i.options) > 0 {
 			if i.selectedIndex == 0 {
 				i.selectedIndex = len(i.options) - 1
@@ -81,6 +85,7 @@ func (i *Input) onRune(config *PromptConfig) terminal.OnRuneFn {
 			i.answer = i.options[i.selectedIndex].Value
 		} else if key == terminal.KeyTab && i.Suggest != nil {
 			i.answer = string(line)
+			i.typedAnswer = i.answer
 			options := i.Suggest(i.answer)
 			i.selectedIndex = 0
 			if len(options) == 0 {
@@ -89,6 +94,7 @@ func (i *Input) onRune(config *PromptConfig) terminal.OnRuneFn {
 
 			i.answer = options[0]
 			if len(options) == 1 {
+				i.typedAnswer = i.answer
 				i.options = nil
 			} else {
 				i.options = core.OptionAnswerList(options)
@@ -101,6 +107,7 @@ func (i *Input) onRune(config *PromptConfig) terminal.OnRuneFn {
 			if key >= terminal.KeySpace {
 				i.answer += string(key)
 			}
+			i.typedAnswer = i.answer
 
 			i.options = nil
 		}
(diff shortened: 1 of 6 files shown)


terminal/runereader.go
```diff
@@ -31,7 +31,13 @@ func (rr *RuneReader) printChar(char rune, mask rune) {
...
 }
 
-func (rr *RuneReader) ReadLine(mask rune) ([]rune, error) {
+type OnRuneFn func(rune, []rune) ([]rune, bool, error)
+
+func (rr *RuneReader) ReadLine(mask rune, onRunes ...OnRuneFn) ([]rune, error) {
+	return rr.ReadLineWithDefault(mask, []rune{}, onRunes...)
+}
+
+func (rr *RuneReader) ReadLineWithDefault(mask rune, d []rune, onRunes ...OnRuneFn) ([]rune, error) {
```
AlecAivazis, line 40: I'm a bit nervous about adding this much logic to a core terminal method. Can you walk me through why the current implementation doesn't work?

lucassabreu: the current input with the suggest allows you to switch between selecting in the list and typing it normally.

if you are typing and press tab, the list appears and works like a select, but if you type a letter it goes back to a normal input.

with the previous implementation using ReadRune directly we did just re-print the "input format" and started printing each rune as it was typed.

the function ReadLine does not allow such control... but if we can set a default input, but keeping the cursor knowing that is a editable section then we can set the input for the user.

it is working like this: https://asciinema.org/a/NjD9zy242lY25NXUreS9J2Lim

```go
package main

import (
	"fmt"
	"os"

	"github.com/AlecAivazis/survey/v2/terminal"
)

func main() {
	rr := terminal.NewRuneReader(terminal.Stdio{
		In:  os.Stdin,
		Out: os.Stdin,
		Err: os.Stderr,
	})
	fmt.Print("some input: ")
	rr.SetTermMode()
	line, err := rr.ReadLineWithDefault(0, []rune("just a start"))

	fmt.Printf("Line: %s, Err: %s", string(line), err)
}
```

i did that way because you changed to use [`ReadLine`](https://github.com/AlecAivazis/survey/pull/361/commits/b52624ad8689c00aeb854fa9469fbf07c8f090e7), I think this is better than adding more ifs for the arrows and duplicating code for the same effect.

lucassabreu: i think it does not support it at least.

AlecAivazis: Ahhh okay so basically the default argument is a way to keep the input in place between tabs - got it! Thanks for explaining

lucassabreu:
@AlecAivazis i think it's done, but there is a error on Ubuntu, you know what it is?

AlecAivazis:
Looks like the test on `survey_test.go:249` is hanging

lucassabreu:
fixed the hanging looks like some problem with using `i` on `vi` in the actions... changed all to `cc`.

it also fixed the compatibility problem with my vim when i did that.

AlecAivazis:
Nice!! I'll find some time today to test this 👍

AlecAivazis (commented):
I pulled down the changes and they behave **perfectly**. It was very satisfying to press the arrow key in the middle of suggestions and then fix the value. Overall a very smooth experience. A little question about the help behavior and then all that should be left is to review the actual code. I've been looking over it periodically and it seems good, just want to give it a final once-over. 

thanks again!

input.go
```diff
@@ -155,27 +161,34 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
...
-		}
+		break
+	}
+
+	i.answer = string(line)
+	// readline print an empty line, go up before we render the follow up
+	cursor.Up(1)
+
+	// if we ran into the help string
+	if i.answer == config.HelpInput && i.Help != "" {
```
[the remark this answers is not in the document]
lucassabreu: @AlecAivazis you mean, if the person send `?` as the response two times in sequence, than we would accept it?

it is not a complex thing to do, i am just thinking that it was a "auto-validation" of the input before... if we change it now, it would break some expectations right?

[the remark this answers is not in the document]
AlecAivazis: That was my thought but i think it's probably better to leave it for another time. No one has brought it up before so it's probably not worth breaking anyone's tests if they are assuming the current behavior.

AlecAivazis (approved):
I think this is good to go! I pulled it down and tested the changes on `iTerm2` for mac , as well as  `cmd.exe` and the `Ubuntu` terminal on windows. Thanks for getting this out so quickly @lucassabreu

AlecAivazis:
This is now published under `v2.2.14`

PR #362, opened 2021-07-03
Validators for the number of responses in multiselect

MAD-py:
In order to have a more complete API and remove worries to the user, I propose to add a series of validators that can make things easier for the user, at this moment I only propose two very simple validators to validate the maximum and minimum of selected answers in a multiselect.

I would like to know what do you think about the proposal and if it is possible to make more validators that can be useful. 

Personally I would like to have a number of validators similar to the [Pydantic](https://pydantic-docs.helpmanual.io) library of python.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,8 @@ validators include:
 | Required     | any         | Rejects zero values of the response type                    | Boolean values pass straight through since the zero value (false) is a valid response |
 | MinLength(n) | string      | Enforces that a response is at least the given length       |                                                                                       |
 | MaxLength(n) | string      | Enforces that a response is no longer than the given length |                                                                                       |
+| MaxItems(n)  | []OptionAnswer | Enforces that a response has no more selections of the indicated |                                                                               |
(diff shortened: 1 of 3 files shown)


AlecAivazis:
Sorry i haven't gotten to this yet @MAD-py. work was really busy this week, will try to find some time to look at it over the weekend

MAD-py:
@AlecAivazis no problem. Actually I had a pretty busy week too.

AlecAivazis (changes requested):
Overall this looks pretty good - thanks for adding some tests. Do you mind adding something in the documentation? Also, I'm open to adding a bunch more validators, they're pretty harmless to throw into a `survey/validators` directory but we can do that in a separate PR

MAD-py:
hi @AlecAivazis 

I've already added the validators info in the READM, I'll be working on more validators that are useful for the API and in a while I'll be doing the corresponding PR.

AlecAivazis:
Thanks for doing that! I'll merge this in once github gives us the 👍

AlecAivazis (approved):

PR #366, opened 2021-07-29
Add WriteAnswer support for promoted fields

### LINKED ISSUE #365: Promoted fields / composition doesn't work
go version go1.16 linux/amd64

```go
type Common struct{
	Name string
}

type Register struct{
	Common // Name field added by composition
	Password string
}
```
```go
[]*survey.Question{
	{
		Name: "Name",
		Prompt: &survey.Input{Message: "Your name"},
		Validate: survey.Required,
	},
	{
		Name:   "Password",
		Prompt: &survey.Password{Message: "Password"},
		Validate: survey.Required,
	},
}
```

I would expect promoted fields to be recognized by the library so redundancy can be avoided for common fields and behavior. 

**What did you see instead?**
Will always result in `could not find field matching Name` because `core/write.go.findFieldIndex()` doesn't check promoted fields.

---

Promoted fields can be checked using the following condition:
```go
field := s.Field(i)
fieldType := sType.Field(i)
if field.Kind() == reflect.Struct && fieldType.Anonymous {
	// This field is a promoted struct
}
```
(Variable names are based on what's used in `core/write.go.findFieldIndex()`.)

System-Glitch:
fixes #365

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -24,6 +24,11 @@ type OptionAnswer struct {
 	Index int
 }
 
+type reflectField struct {
+	value     reflect.Value
+	fieldType reflect.StructField
+}
+
 func OptionAnswerList(incoming []string) []OptionAnswer {
 	list := []OptionAnswer{}
 	for i, opt := range incoming {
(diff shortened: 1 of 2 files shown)


AlecAivazis (approved):
Thanks for this! especially appreciate the extra tests

PR #367, opened 2021-08-03
fix: smkx keypad-transmit mode

### LINKED ISSUE #228: Arrow keys don't work when terminal is in keypad-transmit mode
To reproduce:

1. Start a clean Bash session.

```bash
bash --norc
```

2. Switch terminal to keypad-transmit mode (`smkx`).

```bash
echo $'\e[?1h\E='
```

3. Run `countrylist` example.

```bash
go run examples/countrylist.go
```

4. Press down arrow key.

Expected: Country selection moves down.

Actual: Country selection doesn't move and an error message is printed:

```text
Unexpected Escape Sequence: ['\x1b' 'O']
```

The reason is that in keypad-transmit mode down arrow key sends `^[OB` instead of `^[[B` that `survey` expects.

More info:

- OS: Ubuntu 18.04.
- Terminal: GNOME Terminal.
- Discovered through: https://github.com/romkatv/powerlevel10k/issues/90.

### LINKED ISSUE #341: Missing support for DECCKM cursor keys

https://github.com/cli/cli/issues/3071 has more context, but:

* In any terminal on macos
* Run powershell (it enables DECCKM)
* Now run `gh auth login`
* Press cursor up or down

The cursor keys should navigate to select one of the login options

**What did you see instead?**

could not prompt: Unexpected Escape Sequence: ['\x1b' 'O']

# Analysis

Due to https://github.com/PowerShell/PowerShell/issues/12268 powershell enables DECCKM which changes the cursor key encoding from `CSI A..D` to `SS3 A..D` instead.  SS3 is encoded as `ESC O` in 7-bit encoding.

It is technically a bug that powershell leaves the cursor keys in that mode.
However, it should be pretty simple to support this encoding here in survey:

https://github.com/AlecAivazis/survey/blob/dbdb3e56a5e332b0478e0a8e8522e53a49fb7dcd/terminal/runereader_posix.go#L79-L81

by allowing either `[` or `O`

ldez:
> To see the difference between normal/application modes, consider this example:
>
> - In normal (non-application) mode, the terminal transmits a down-arrow as \E[C, which happens to echo as a down-arrow.
> - In application mode the terminal transmits \EOC, which echoes as C. That is because the \EO is the SS3 control, which says to use the character from the G3 character set for the next cell.

https://invisible-island.net/xterm/xterm.faq.html#xterm_arrows

> ESC =     Application Keypad (DECKPAM).
>
> ESC >     Normal Keypad (DECKPNM), VT100.

https://invisible-island.net/xterm/ctlseqs/ctlseqs.html

See https://vt100.net/docs/vt102-ug/appendixc.html

Fix #228
Fix #341

### DIFF
--- a/terminal/runereader_posix.go
+++ b/terminal/runereader_posix.go
@@ -1,4 +1,3 @@
-//go:build !windows
 // +build !windows
 
 // The terminal mode manipulation code is derived heavily from:
--- a/terminal/runereader_posix.go
+++ b/terminal/runereader_posix.go
@@ -1,3 +1,4 @@
+//go:build !windows
 // +build !windows
 
 // The terminal mode manipulation code is derived heavily from:
@@ -14,6 +15,11 @@ import (
 	"fmt"
 	"syscall"
 	"unsafe"
+)
+
+const (
+	normalKeypad      = '['
+	applicationKeypad = 'O'
 )
 
 type runeReaderState struct {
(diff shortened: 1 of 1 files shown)


terminal/runereader_posix.go
```diff
@@ -105,7 +105,6 @@ func (rr *RuneReader) ReadRune() (rune, int, error) {
 			rr.state.reader.Discard(1)
 			return IgnoreKey, 1, nil
 		}
-		return r, size, fmt.Errorf("Unknown Escape Sequence: %q", []rune{'\033', '[', r})
```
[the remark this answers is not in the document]
mislav: Ah, you are right!

terminal/runereader_posix.go
```diff
@@ -76,7 +76,7 @@ func (rr *RuneReader) ReadRune() (rune, int, error) {
 		if err != nil {
 			return r, size, err
 		}
-		if r != '[' {
+		if r != '[' && r != 'O' {
```
adonovan, line 79: While you're here, you might want to:
- add a link to a terminal reference such as https://vt100.net/docs/vt102-ug/appendixc.html (for VT100).
- comment on each of the switch cases what sequence it is intended to recognize
  e.g. `case 'A': // ESC [ A or ESC O A`
- reject illegal sequences such as 'ESC O 3'. That means (in effect) using different switch cases depending on whether the prefix was ESC [ or ESC O.
- lowercase the error message (and put the word "terminal" in there): "unexpected escape sequence from terminal: %q"

I confirm that this PR fixes the problem I referred to in https://github.com/AlecAivazis/survey/issues/228#issuecomment-915493711. Thanks!

terminal/runereader_posix.go
```diff
@@ -1,3 +1,4 @@
+//go:build !windows
```
mislav, line 1: Even though this is a newer syntax of the build directive, I do not think it's needed as it doesn't offer any improvements over `// +build !windows` and is only supported since Go 1.17.

mislav (approved):
Loving the new verbose, documented approach!

terminal/runereader_posix.go
```diff
@@ -76,36 +81,126 @@ func (rr *RuneReader) ReadRune() (rune, int, error) {
...
 			return r, size, err
 		}
-		if r != '[' {
-			return r, size, fmt.Errorf("Unexpected Escape Sequence: %q", []rune{'\033', r})
-		}
-		r, size, err = rr.state.reader.ReadRune()
-		if err != nil {
-			return r, size, err
-		}
+
```
adonovan, line 84: What I had in mind wasn't quite so verbose---only one of the switches varies based on O vs [.
Also, a single doc link is enough. Readers can search the docs if they need to.

How about something more like this:

```
func (rr *RuneReader) ReadRune() (rune, int, error) {
	r, size, err := rr.state.reader.ReadRune()
	if err != nil {
		return r, size, err
	}

	// ANSI terminal controls.
	// See https://vt100.net/docs/vt102-ug/appendixc.html
	const (
		escape            = '\x1b'
		normalKeypad      = '['
		applicationKeypad = 'O'
	)

	// Parse escape sequences such as ESC [ A for arrow keys.
	if r == escape {
		if rr.state.reader.Buffered() == 0 {
			// no more characters so must be Escape key
			return KeyEscape, 1, nil
		}
		r, size, err = rr.state.reader.ReadRune()
		if err != nil {
			return r, size, err
		}

		// ESC O ... or ESC A ...?
		var normal bool
		switch r {
		case normalKeypad:
			normal = true

		case applicationKeypad:
			normal = false

		default:
			return r, size, fmt.Errorf("unexpected escape sequence from terminal: %q", []rune{'\033', r})
		}

		r, size, err := rr.state.reader.ReadRune()
		if err != nil {
			return r, size, err
		}

		switch r {
		case 'A': // ESC [ A or ESC O A
			return KeyArrowUp, 1, nil
		case 'B': // ESC [ B or ESC O B
			return KeyArrowDown, 1, nil
		case 'C': // ESC [ C or ESC O C
			return KeyArrowRight, 1, nil
		case 'D': // ESC [ D or ESC O D
			return KeyArrowLeft, 1, nil
		case 'F': // ESC [ F or ESC O F
			return SpecialKeyEnd, 1, nil
		case 'H': // ESC [ H or ESC O H
			return SpecialKeyHome, 1, nil
		case '3': // ESC [ 3
			if normal {
				// discard the following '~' key from buffer
				_, _ = rr.state.reader.Discard(1)
				return SpecialKeyDelete, 1, nil
			}
		}

		// discard the following '~' key from buffer
		_, _ = rr.state.reader.Discard(1)
		return IgnoreKey, 1, nil
	}
}
```

ldez: sorry but you are saying the opposite of @mislav, I will apply your suggestions but it's a bit difficult for me to don't have a unique voice or a maintainer of this project on this topic.

adonovan: Sorry for not being more precise earlier. BTW, neither @mislav nor I are maintainers of this repo, so you don't have to listen to anything we say. :) @AlecAivazis is the authority in these parts.

adonovan (approved):
LGTM. Thanks!

AlecAivazis:
Fantastic - thanks for getting this through everyone. 

Sorry  i wasn't able to contribute much to the conversation but it seems you guys got to a reallly nice spot. I'll merge this once the actions gives us a 👍

AlecAivazis (approved):

PR #372, opened 2021-09-01
Fix build on Solaris/illumos.

jperkin:
Fill in the missing ioctlReadTermios/ioctlWriteTermios, and switch to
syscall.Syscall as Syscall6 is not available.  Probably makes this work
on AIX too as a bonus.

### DIFF
--- a/terminal/runereader_posix.go
+++ b/terminal/runereader_posix.go
@@ -44,7 +44,7 @@ func (rr *RuneReader) Buffer() *bytes.Buffer {
 
 // For reading runes we just want to disable echo.
 func (rr *RuneReader) SetTermMode() error {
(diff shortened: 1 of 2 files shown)


mislav (commented):
Thanks; this looks good!

What do you think of https://github.com/AlecAivazis/survey/pull/377 as an alternative to this approach? It feels like runereader largely just replicates what Go's own `x/term` package is doing, so to increase portability and to avoid having to maintain this many build tags, we could just offload it to the external package.

jperkin:
Yeh that sounds great, happy to test that out once it's merged.

jperkin:
Hi, any update on this?  It's a bit of a pain having to merge these changes every time influxdb switches to a different version of survey.  Thanks!

PR #389, opened 2021-11-08
fix: error message for MinItems validator

mostafahussein:
Currently when using `survey.MinItems(1)` with no items selected it says:
```
X Sorry, your reply was invalid: value is too long. Min items is 1
```
However it should be `too short` not `too long`


mislav (approved):
Good catch!

Issue #416, opened 2022-04-20
Option to add comment/description on select option


luanraithz:
I created a [fork](https://github.com/luanraithz/survey) adding a option to show a description for a given select's option.

You can see how it looks [here](https://github.com/luanraithz/lvb)

If you guys are interested on that, tell me so I give a refact and open a pull request.

Cheers!

AlecAivazis:
Hey @luanraithz! I think it looks like a very interesting addition - thanks for offering to submit a PR. I'd gladly take a look 👍

mislav:
Thank you! https://github.com/AlecAivazis/survey/commit/c2be27f4a7faa58a920ea402d0984647a5b8deb2

PR #418, opened 2022-04-27
Added an option to add a comment to every select's option

luanraithz:
This is the pull request mentioned on #416


AlecAivazis (approved):
Straight forward implementation and works as expected - nice!

Will merge this in once we get the 👍 from GH

AlecAivazis:
@luanraithz the tests are failing - mind fixing them?

luanraithz:
Sorry, didn't see that. I'm on it

README.md
```diff
@@ -185,12 +185,23 @@ The user can also press `esc` to toggle the ability cycle through the options wi
 By default, the select prompt is limited to showing 7 options at a time
 and will paginate lists of options longer than that. This can be changed a number of ways:
 
-```golang
-// as a field on a single select
-prompt := &survey.MultiSelect{..., PageSize: 10}
+#### Select option's description
 
-// or as an option to Ask or AskOne
-survey.AskOne(prompt, &days, survey.WithPageSize(10))
```
mislav, line 193: Why does this remove the WithPageSize example? Can the previous documentation also stay?

README.md
```diff
@@ -185,12 +185,23 @@ The user can also press `esc` to toggle the ability cycle through the options wi
...
+![Example](img/description-demo.gif)
```
mislav, line 190: I appreciate the demo gif, but it would have been enough to just share it in the body of your PR. I don't think it should be present in the git repository itself since this is a Go library that gets downloaded every time it gets included into a project, and it would be great if we could avoid downloading 85kB of unneeded data.

examples/meals.go
```diff
@@ -0,0 +1,53 @@
+package main
```
mislav, line 1: Can the file `meals.go` be named something more descriptive of the feature it's demoing, such as `select_description.go`?

examples/meals.go
```diff
@@ -0,0 +1,53 @@
...
+		{
+			Name: "meal",
+			Prompt: &survey.Select{
+				Message: "Choose a meal:",
+				Options: titles,
+				Description: func(value string, index int) string {
+					return meals[index].Comment
+				},
+			},
+			Validate: survey.Required,
```
mislav, line 40: This isn't a demo for validation features, and since validation isn't necessary here, we could just omit this.

examples/meals.go
```diff
@@ -0,0 +1,53 @@
...
+		},
+	}
+
+	// ask the question
+	err := survey.Ask(qs, &answers)
```
mislav, line 45: Could this use AskOne since there is only one question being asked? I think it would result in simpler code.

Additionally, the destination for the answer could be a pointer to an int. That way, survey will write the index of the selected feature upon completion. I think the index would be more useful for the user, since they can query the original `meals` slice with it.

mislav (changes requested):
I love this feature! Thanks for adding it.

README.md
luanraithz: Sorry, no idea why I did that

luanraithz:
I think it's all solved, thanks for the review!

mislav (approved):
Thank you!

dvaghela:
@AlecAivazis How soon can we expect a release with this PR?
And thanks for the project, it is really great! 😄

AlecAivazis:
Thanks for the ping! It's now been released as `2.3.5`

Issue #412, opened 2022-03-05
Get Selected Index in AskOne


jojomi:
I would like to extend survey for my purposes, actually making it more expressive using generics.

The idea is to be able to select a value from any type of slice without type assertions using a mapper func that returns a displayable string for every element of the slice:

``` Go
func ChooseOneWithMapper[T any](question string, options []T, mapper func(t T) string) (T, error) {
	var tResult T

	// map options
	stringOpts := make([]string, len(options))
	for i, option := range options {
		stringOpts[i] = mapper(option)
	}

	prompt := &survey.Select{
		Message: question,
		Options: stringOpts,
	}

	var result string
	err := survey.AskOne(prompt, &result, nil)
	if err != nil {
		return tResult, err
	}

	// map back
	for i, option := range stringOpts {
		if option != result {
			continue
		}
		return options[i], nil
	}
	return tResult, fmt.Errorf("invalid selection: %s", result)
}
```

This is not perfect though because I have to map the selection back to the original slice which may or may not be correct. It would be reliably possible if I could get the *index* of the selection instead of the value. Is there a way to do so I overlooked in the current code?

mislav:
If `&result` is a pointer to an `int` instead of a `string`, it will write the index of the selection instead of the text of the selection.

jojomi:
Thank you, never thought about that!

jojomi:
I tried to make us of that but I think it is buggy:

``` go
package main

import (
	"fmt"

	"github.com/AlecAivazis/survey/v2"
)

func main() {
	var answer int

	prompt := &survey.Select{
		Message: "Choose a color:",
		Options: []string{"yellow", "red", "red", "blue"},
	}
	survey.AskOne(prompt, &answer)

	fmt.Printf("answer index returned: %d\n", answer)
}
```
When selecting the answers in separate runs, it returns 0, 2, 2, 3 respectively, but I would expect to see 0, 1, 2, 3. It seems buggy due to two answers sharing the same description string.

mislav:
> It seems buggy due to two answers sharing the same description string.

I think you are right. I will reopen to track returning the right index, even if options share the same string. Note that from the user's perspective, there should never be options with identical strings, otherwise the user can't know which one is "correct" to select.

jojomi:
You are right about the UX concerns, but also about fixing it, since it totally seems possible to return the technically correct value here. Thank you already!

PR #420, opened 2022-05-12
Ensure accurate selected index returned from Select

mislav:
Fixes https://github.com/AlecAivazis/survey/issues/412 which describes the case where the selected index would be mis-reported when some options happen to share identical value. Since the Select component internally tracks the `selectedIndex`, just use that value instead of trying to determine it by comparing strings.

@AlecAivazis In my attempts to clean up default value handling here, I fear I have introduced backwards-incompatible changes. I could restore the previous behavior, sure, but I wanted to ask if it's also worth "fixing" the default value handling. Namely:

- Before, when a string value was passed, for example `Default: "moo"`, this default was used in the result when the user immediately pressed <kbd>Enter</kbd> (i.e. without otherwise interacting with the Select component) **even if the value `moo` did not exist among Options**. Now, **the prompt will error out** noting that the default value must be present among Options, mainly so that it can be visually pre-selected when rendering the widget.

- Before, when an invalid int value was passed, for example `Default: 1000` when there are fewer than 1000 Options, a panic would happen when the user immediately pressed <kbd>Enter</kbd>, _but not otherwise_. Now, the prompt will **immediately error out** noting that the Default value is invalid since it's out of bounds.

So, what I changed here felt like fixing bugs to me, but I can imagine that some users might be relying on these features, for example having a Default that's not among Options so they can detect when the user wanted to skip a Select. I'm not sure if those features were worth keeping, though. For example, consider this prompt:

```
? Choose a country:  [Use arrows to move, type to filter]
> Afghanistan
  Åland Islands
  Albania
```

If I press <kbd>Enter</kbd> here, I would expect that "Afghanistan" got selected. However, if there was a Default value such as "moo", that string will be used instead of "Afghanistan", which I find counter-intuitive. I would rather break this feature than have users rely on it.

What are your thoughts?

### DIFF
--- a/select.go
+++ b/select.go
@@ -2,6 +2,7 @@ package survey
 
 import (
 	"errors"
+	"fmt"
 
 	"github.com/AlecAivazis/survey/v2/core"
 	"github.com/AlecAivazis/survey/v2/terminal"
@@ -30,7 +31,6 @@ type Select struct {
 	Filter        func(filter string, value string, index int) bool
 	filter        string
 	selectedIndex int
-	useDefault    bool
 	showingHelp   bool
 }
 
@@ -94,8 +94,6 @@ func (s *Select) OnChange(key rune, config *PromptConfig) bool {
 
 		// if the user pressed the up arrow or 'k' to emulate vim
 	} else if (key == terminal.KeyArrowUp || (s.VimMode && key == 'k')) && len(options) > 0 {
-		s.useDefault = false
-
 		// if we are at the top of the list
 		if s.selectedIndex == 0 {
 			// start from the button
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Oh, man. This is one of those super tough situations. I agree with you that these are really bug fixes but it seems very likely that someone out there is relying on the silent failure or doesn't know they have an invalid default value. 

I'm leaning more towards keeping this under `v2` and we should just make sure its very clear what happened in the release notes.

AlecAivazis (commented):
Changes look good! I think it's probably worth adding some test cases to the select test file so we make sure this behavior stays in place

AlecAivazis:
@mislav i'm going to merge this since it's been around for so long and we both agree it's an appropriate change

mislav:
@AlecAivazis Thanks! Do you still think this should be a v2? Strictly speaking, I think it should.

skelouse:
This is a backwards incompatible change for me, as if no default I'm sending an empty string.  Although I can fix it, others may be strongly relying on this functionality

Issue #196, opened 2019-05-17
Allow editing of existing text


syvanpera:
I couldn't find a way to edit some existing text using survey. You can set the Default text for a prompt, which will show it as a grey text next to the prompt and is used if no input is given. But I'd like to be able to edit this default text and save that when enter is pressed.

syvanpera:
Actually now that I think of it maybe the cleanest way to do this would be to display the text set in the response initially.
So basically:
`name := "Some editable text"
prompt := &survey.Input{
    Message: "ping",
}
survey.AskOne(prompt, &name, nil)`

would display a prompt with "Some editable text" in it already

MarkusFreitag:
Yes correct, that would be the recommended way to achieve such "dynamic" input components.

AlecAivazis:
I'm going to close this since its been resolved. Thanks again for following through with this @MarkusFreitag

syvanpera:
Umm, so are you saying this should work? Because at least the example I gave, does not

AlecAivazis:
Ah sorry! I was under the impression @MarkusFreitag had tested it and verified it worked. I'm going to reopen this ticket and assign it a `Feature` tag.

syvanpera:
Ah okay. I might be willing to explore if I can implement this. I'm just a bit new with the different ways go modules are built as I've only made my own modules using the new go module feature and I'm not quite sure how to start contributing to for example this project. Can you give me a quick run down on how to build and test this project?

AlecAivazis:
Awesome! I'd love to help you anyway I can. This project uses [run](https://github.com/AlecAivazis/run) to handle various tasks so once you have downloaded that with `go get github.com/alecaivazis/run` you can run the tests with `run tests`. Since you won't have to touch any packages that are outside of `survey` you shouldn't need to do anything fancy for gomodules to work.

Also, if you prefer or want more immediate feedback, feel free to reach out on the gopher slack channel!

syvanpera:
Okay, cool. I got run installed, but the tests are failing. Should they pass currently? Do you have own channel for this project on gophers slack?

AlecAivazis:
Nope, you can just DM me 😄

They should be passing (travis has a green light). Wanna send me some screenshots of what you're seeing?

Wulfheart:
> would display a prompt with "Some editable text" in it already

I disagree. It just shows "ping".

Wulfheart:
[...]

PR #421, opened 2022-05-12
Allow editable default for Input prompt

mislav:
When doing
```golang
survey.Input{
	Message: "What is your name?",
	Default: "Jordan",
}
```
the default value "Jordan" is now editable and can be navigated around with the cursor if `survey.WithShowCursor(true)` was used.

Promoting visibility for this idea: https://github.com/AlecAivazis/survey/issues/196#issuecomment-1067042464

@AlecAivazis Since this is backwards incompatible, what do you think about adding a separate field for editable default value? E.g. `DefaultEditable: "Jordan"`

### DIFF
--- a/examples/cursor.go
+++ b/examples/cursor.go
@@ -12,8 +12,8 @@ var simpleQs = []*survey.Question{
 		Name: "name",
 		Prompt: &survey.Input{
 			Message: "What is your name?",
+			Default: "Jordan",
 		},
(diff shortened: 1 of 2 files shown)


AlecAivazis:
Hm, good question. I think having a new field is "alright". it's definitely not great but we're kind of stuck with it. If there are enough little breaks maybe its time to get them together?

Issue #428, opened 2022-05-31
Feature Request: Survey Mock


SirRegion:
Hi @AlecAivazis,
thank you for creating this great library.

I used it a lot in my work project, but I have now reached a point, where I spend a lot more time debugging than implementing features. The main reason: missing tests 😢.

So I started adding tests and implemented wrappers and fakes for all my os calls (or used a library that included those). 

Survey is now the only dependency with side effects that I use that has no wrappers / mocks available. 
I tried writing them myself, but I got really lost when trying to correctly implement a mock that still correctly sets the responses I have given it. (Reflection is really not my strong suit 😆)

So my proposal would be the following:

### 1. Add a survey interface and struct

Adding a interface that defines all the functions survey provides and a struct that implements all those functions would enable the user to pass around the survey struct and use dependency injection instead of hardcoding access to the survey package.

It would also allow users to extend survey without having to wait until their pull request is merged into this repo.
This would also allow storing configurations inside the struct, removing the requirement to pass them to each Ask or AskOne function.

### 2. Add a survey mock struct

Adding a survey mock struct that implements the interface, but does not actually show any output on the command line would enable users to create unit tests without having to rely on a virtual terminal. Not only would this speed up tests, but also improve their clarity.

For Example: You want to test your newly written function `askRandomQuestion` asks the predefined question "Are you there?" and if the user responds with "Yes". If you're using `go-expect`, you first have to set up the environment, pass stdio and stderr to survey and than carefully check the complete text on the terminal, to make sure everything is working correctly.

But if you would have access to a mock, you could simply call `surveyMock.getLatestQuestion` to confirm if the latest question really was "Are you there". And if you also want to check the result of the function, you cound simply set the response by calling `surveyMock.setResponse("Yes")` before executing the function and then confirm afterwards.

I know this would probably be a lot of work, but I would really love to have this feature in survey!

PS: I would be open to try to implement the first part, but that change does not make sense if the second part is not being considered. (And the second part would have to be someone else's responsibility, I don't know how to do it 😁)

AlecAivazis:
[...]

PR #429, opened 2022-06-02
Added an interface for all the public functions of the survey package

SirRegion:
This PR adds an interface for survey that can be used instead of calling the functions directly.

It also includes a struct with no contents hat implements the interface.

This lays the groundwork for adding a survey mock as discussed in issue #428.

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,11 +456,9 @@ For some examples, you can see any of the tests in this repo.
 
 ### Using a mock (unstable API)
 
-#### Warning: 
-
+>**Warning:** 
 > The Mock API is currently still unstable and subject to change. 
 > Once it's done, it will be the recommended way to test survey. 
-
 > If you are unsure what to use right now, use a simulated terminal.
 
 Instead of calling the survey functions directly, you can create a survey struct and call the functions from there.


mislav:
Thanks for taking this on! Since this is a work in progress, would you mind marking the PR as draft until it's ready for review?

SirRegion:
Sure!

mislav:
@SirRegion I meant, instead of changing the title of the PR to include the word "draft", actually making the PR as draft in GitHub web UI. I believe that the option to do it is in the right sidebar for the PR.

SirRegion:
Oh, thank you. I'm way more familiar with GitLab than Github, where adding "draft:" to an PR automatically converts it. I fixed it now :)

SirRegion:
I added an implementation for the mock (including tests) and some documentation in the readme on how to use it. 

You can now start reviewing and merging if you think it's ready :)

surveyMock.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/AlecAivazis/survey/v2/terminal"
+)
+
+type SurveyMock struct {
+	singleResponse    interface{}
+	multipleResponses map[string]interface{}
+}
+
+// Not implemented, because it does not affect the mock
+func (mock *SurveyMock) WithStdio(in terminal.FileReader, out terminal.FileWriter, err io.Writer) AskOpt {
```
mislav, line 16: Is there any reason to keep these options funcs at the Survey interface level? Calling them as part of the interface has no utility: they can't be mocked, they can't be verified, and they only forward the call to the funcs on the package level. Could we just keep them on the package level instead?

README.md
```diff
@@ -450,6 +454,65 @@ stdio to an in-memory [virtual terminal](https://github.com/hinshun/vt10x).
...
+
+#### Test
+
+```golang
+func TestAskForPie(t *testing.T) {
+    
+    //create mock
+    mock := survey.SurveyMock{} 
+    //set the response the "user" should select 
+    mock.setResponse(true)
```
mislav, line 504: If AskForPie is going to call `survey.Ask()` multiple times, how would we mock all those using `SetResponse`?

README.md
```diff
@@ -450,6 +454,65 @@ stdio to an in-memory [virtual terminal](https://github.com/hinshun/vt10x).
...
+
+```golang
+
+survey := survey.Survey{}
+
+response := false
+prompt := &survey.Confirm{
+    Message: "Do you like pie?",
+}
+survey.AskOne(prompt, &response)
```
mislav, line 469: Now the big question is whether we should amend all other examples in the Readme to call `Ask` and `AskOne` on the Survey struct instead of on the package level? That would make everyone's code ultimately testable by default. Otherwise, the Readme will offer users usage examples of survey on the package level, but later the testing section will effectively tell them they need to rewrite their original code if they hope to mock Ask/AskOne calls. My hope for Survey is that it's testable by default without the need of changing the original implementation. @AlecAivazis Thoughts?

README.md
```diff
@@ -450,6 +454,65 @@ stdio to an in-memory [virtual terminal](https://github.com/hinshun/vt10x).
 
 For some examples, you can see any of the tests in this repo.
 
+### Using a mock
+
+Instead of calling the survey functions directly, you can create a survey struct and call the functions from there.
+
+```golang
+
+survey := survey.Survey{}
```
mislav, line 463: This shadows the `survey` package with a `survey` variable. It's perfectly valid Go, but it might confuse the reader, since past this point `survey` isn't what it was before. For clarity, could we call the struct something different? I'm also thinking whether there would be a better name for the Survey struct (for the same reason)

README.md
```diff
@@ -450,6 +454,65 @@ stdio to an in-memory [virtual terminal](https://github.com/hinshun/vt10x).
...
+    
+    //create mock
+    mock := survey.SurveyMock{} 
+    //set the response the "user" should select 
+    mock.setResponse(true)
+
+    result := AskForPie(mock)
+    
+    //check output of the function
+    if result != true {
```
mislav, line 509:
```suggestion
    if !result {
```

README.md
```diff
@@ -450,6 +454,65 @@ stdio to an in-memory [virtual terminal](https://github.com/hinshun/vt10x).
...
+        t.Fatalf("AskForPie returned false, but it should have returned true!")
```
mislav, line 510:
```suggestion
        t.Fatal("AskForPie returned false, but it should have returned true")
```

mislav (changes requested):
This looks promising! Thank you for working on it.

Over in the GitHub CLI codebase we have experience with mocking Survey using our own approach. One thing we learned is that our tests are more resilient over a longer period of time if we mock Survey in a way that the mocks are still valid even if the underlying implementation switches from Ask to AskOne (and vice versa). Also, we have ability to mock multiple sequential calls of Ask/AskOne. We do this by matching result stubs not to Ask/AskOne calls, but [to prompt messages](https://github.com/cli/cli/blob/trunk/pkg/cmd/issue/create/create_test.go#L405-L409). What do you think of this approach, and would you be open if I helped port it over to what you are proposing in your PR?

README.md
SirRegion: I explicitly added this comparison for clarity for less experienced developers, but the short form is probably understandable enough as well. 

I'll leave the decision up you.

README.md
SirRegion: I noticed this while writing, but i couldn't come up with a better name :(

Any suggestions?

README.md
SirRegion: I guess updating the readme should not be to much work, as it's just one or two lines of code changed in each example.

I left it the same for this pull request, because I wasn't sure how much of a "breaking" change you would accept for the docs.

README.md
SirRegion: The current implementation does not allow to set multiple responses and would always use the last one that is set. 
However, I feel like this could be acceptable behaviour, because it forces the developers to separate different `ask` calls into multiple functions that only ever do one thing (as they should 😉)

Alternatively, `setResponse()` could return the struct and store the responses in a slice. This would make this call joinable and allow multiple answers.

surveyMock.go
SirRegion: There are pros and cons to keeping all the functions in the interface:

Pro: 
- Less confusing to have to call some functions on the struct and some on the package#
- Allows for stateful functions in the future

Con:
- It convolutes the library code and forces us to maintain a lot of function declarations in three places instead of two

I'm not sure which aspect is more important, but I would gravitate to keeping them in the interface. 
However, you have the final word 😄

SirRegion:
> This looks promising! Thank you for working on it.
> 
> Over in the GitHub CLI codebase we have experience with mocking Survey using our own approach. One thing we learned is that our tests are more resilient over a longer period of time if we mock Survey in a way that the mocks are still valid even if the underlying implementation switches from Ask to AskOne (and vice versa). Also, we have ability to mock multiple sequential calls of Ask/AskOne. We do this by matching result stubs not to Ask/AskOne calls, but [to prompt messages](https://github.com/cli/cli/blob/trunk/pkg/cmd/issue/create/create_test.go#L405-L409). What do you think of this approach, and would you be open if I helped port it over to what you are proposing in your PR?

Thanks for your suggestions. I basically just build an MVP that is enough for my purposes, but I would be happy to see you implement a more sturdy variant on top of the work I already did. 
I think it would be great to have both the ability to define the response for Ask/AskOne directly and also have the more complex way of returning an answer based on the prompt given.

README.md
mislav: I think it does more of a service to inexperienced developers to show them code that reflects the idiomatic style for that language. I appreciate that you tried to add clarity to documentation examples, but I do not think that logical negation operator will pose a barrier to understanding Survey.

README.md
SirRegion: How about "Surveyer"? Keeps the name of the library but is still not exactly the same.

mislav: That sounds great

SirRegion: I implemented it, but changed to "Surveyor" (does not look like a spelling mistake that way 😄)

README.md
SirRegion: I updated the  examples in the readme, but I did not touch the text for now.

Some parts ("Running the Prompts", "Testing") likely still need updating.

surveyMock.go
SirRegion: I thought about it some more. 

I think the two pros I listed are either weak arguments (number 1) or can still be implemented later, even if we don't add them now (number two). 

I'll remove all the functions except Ask/AskOne from the interface.

README.md
mislav: I feel strongly that the mock solution should be flexible enough to mock all responses, no matter how many times Ask or AskOne has been called. I may be able to contribute this improvement to your mock implementation.

Example use case:
```
surveyer := survey.SurveyMock{}
surveyer.SetReponse(map[string]interface{}{
  "name": "foo",
  "message": "hello",
})
// ...
DoSomething(surveyer)
```
This test setup will work as long as DoSomething is calling Ask with two questions. However, let's say that the implementation of DoSomething needs to change because the prompt of question 2 needs to dynamically adjust to the response from question 1. The way to do that with survey is to switch from Ask to 2x AskOne. After the implementation changes in this way, the current test from the example will break. However, if the mock solution does not support stubbing several AskOnes, then the user will be forced to break up DoSomething into separate functions just to appease the mock implementation. I think this would cause frustration to the user.

SirRegion: This is a usecase i did not think of before.

How would you feel about something like this (inspired by [jest](https://jestjs.io/docs/mock-functions#mock-return-values)):
```golang

mock := survey.Surveyor{}
mock.SetResponseOnce("first").SetResponseOnce("second").SetRespose("all others")

DoSomething(surveyor)

```
Now the mock would respond to the first question inside `DoSomething` with "first", two the second with "second" and to all other questions with "all others". 

Now if `DoSomething` is using `Ask`, it would just fill the question struct with the answers "first", "second", "all others", "all others" if the prompt had four questions. 

If `DoSomething` is using `AskOne`, it would respond to the first `AskOne` with "first", and to the second with "second", etc...

mislav: I like your proposal; it looks more flexible already! One thing that I still think would be great if SurveyMock could do is verify that the prompt was shown to the user with the right text, since this text is the most important part of how the user interacts with Survey. [We do it like this](https://github.com/cli/cli/blob/7d0b37d8f37b554e3acbc4d00c72d24e08be9e90/pkg/cmd/pr/merge/merge_test.go#L1103-L1106), which feels similar to what you've proposed, but allows for a greater level of precision.

SirRegion: I like that way. Can you implement it?

mislav: Yeah! I'm thinking to build open your PR later this week.

SirRegion: Would you merge the PR if I'd clean up the documentation and mark the (new) testing part with a disclaimer that the mock API is still subject to change?

I would like to use the current state in my project and I don't mind having to redo some tests with the new and improved API.

AlecAivazis: I think if we're going down this path we probably also want to provide methods on the mock for setting a specific field with a value (something like `.setFieldResponse("name", "SirRegion")`. I'm not sure there isn't a lot of value in coupling a test to the order in which the questions are asked but I think if someone wants to structure there test that way, we should support them

README.md
AlecAivazis: Hm, good question. I think keeping the examples as short as possible is probably the best. I'd like to avoid the `surveyor := survey.Surveyor{}` at the start of every example so either we can reference an undefined `surveyor` in the examples or we can just use the package-level functions. Either way, we definitely want to add a section about these two different approaches in the "Running the Prompts" section so it shouldn't be too confusing if we _do_ decide to update the tests to reference `surveyor.Ask` without explicitly showing its declaration.

SirRegion: I think explicitly using `surveyor.Ask` in the examples is the way to go, because it makes the code testable by default (like @mislav noted). 

Referencing an undefined `surveyor` is probably not even a bad move, because it might move users directly to passing it to functions instead of creation a new instance everytime they want to ask a question.

I would recommend updating the first example to include a function where the surveyor is passed as an argument to show how it is done "correctly".

README.md
mislav: @SirRegion Yes, let's undo the overall README changes for now (to keep things as they were before) and have there just be a small mention & example of the mock functionality.

@AlecAivazis How would the `SetFieldResponse()` functionality allow for asserting the order in which questions were asked?

AlecAivazis: I mean to have both methods so a user could decide if they want to test the order of responses or just if a response was given for a field.

mislav: @AlecAivazis Agreed! It's probably better to be strict with verification out-of-the-box (i.e. to verify the order of prompts by default) but give the developer a chance to opt out if needed.

SirRegion: @mislav I reverted the commit that changed the readme for now and updated the Testing part to include a warning about the unstable mock API.

Issue #386, opened 2021-10-15
Password input without revealing password length


quiquelhappy:
Hello,
I'd link to render "", instead of "*" for the password input: similar to what linux does with pretty much every password input (aka invisible password input)

AlecAivazis:
This can now be accomplished by setting the `HideCharacter` to a [zero-width space](https://zerowidthspace.me/)

szh:
I tried this today and the cursor moves ahead which still reveals the password length.

szh:
OK. I'll submit a PR for this.

PR #431, opened 2022-06-23
Added customisable hide character for password inputs.

SirRegion:
Hi @AlecAivazis and @mislav!

I had some free time so I tried my hand on implementing #386. 

Changing the hide character of the password field works as expected. However, using an empty rune just results in the characters not getting hidden at all.

The best workaround for this is setting the custom hide character to an [zero width space](https://zerowidthspace.me/). This results in the behaviour required by the original poster.

I included an example for the password input, but I did not write a test for the behaviour. I also noticed that the existing password tests don't actually validate that the input is hidden, just that is is stored correctly. I'm sadly incapable of fixing those two issues.

### DIFF
--- a/examples/password.go
+++ b/examples/password.go
@@ -1,0 +1,36 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/AlecAivazis/survey/v2"
+)
+
+// the questions to ask
+var defaultPasswordCharacterPrompt = &survey.Password{
+	Message: "What is your password? (Default hide character)",
+}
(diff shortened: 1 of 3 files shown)


AlecAivazis:
Thanks for putting this together @SirRegion! Mind putting a little note in the docs about the zero space width?

AlecAivazis:
@SirRegion there is now a breaking change. I'd be happy to merge this once its ready (I can add the documentation after)

Issue #393, opened 2021-12-08
Remove "select all" option from Multiselect


ces131:
Hi there!

For our use case, we want to prevent users from being able to select "all options" (using right right arrow) for Mulitselect. It seems like this could be easily done by adding a field to the `PromptConfig` struct that defaults to still give the "Select All" option, but can be toggled to turn it off (and the template can be updated accordingly as well)

I was going to patch the package but thought that this may be a valid use case for other users as well. I can put up a pull request for it, but wanted to see whether this was of interest :) 

Thanks!!

mislav:
I'm interested. Thank you for the feature idea! How would you envision this to be configured?

ces131:
I think it falls sort of in the same category as the `KeepFilter` config (I noticed it also only applies to `MulitSelect` also) and could be implemented in pretty much the same way - I chose `RemoveSelectAll` for my patch, but we could just make this `SelectAll` or something and set the default to true (just wouldn't follow the "configs being false by default" pattern that's there is so far):
```
type PromptConfig struct {
	PageSize     int
	Icons        IconSet
	HelpInput    string
	SuggestInput string
	Filter       func(filter string, option string, index int) bool
	KeepFilter   bool
	ShowCursor   bool
	RemoveSelectAll bool
}
```

```
// WithRemoveSelectAll - removes the "select all" option in Multiselect
func WithRemoveSelectAll(RemoveSelectAll bool) AskOpt {
	return func(options *AskOptions) error {
		// set the page size
		options.PromptConfig.RemoveSelectAll = RemoveSelectAll

		// nothing went wrong
		return nil
	}
}
```
(and set the default to false in the `defaultAskOptions`)

Then in multiselect, we can just skip the `KeyRight` binding with:
```
} else if !config.RemoveSelectAll && key == terminal.KeyArrowRight {
		for _, v := range options {
			m.checked[v.Index] = true
		}
		if !config.KeepFilter {
			m.filter = ""
		}
```

and remove it from the template with:
```
... space to select, {{- if not .Config.RemoveSelectAll }}<right> to all,{{end}} <left> to none, ...
```

mislav:
That sounds like a great plan. I would encourage you to submit a PR, since you've already identified all the places that need the change 👍

AlecAivazis:
Closing this since #439 has been merged

PR #439, opened 2022-08-05
Add a RemoveSelectAll and RemoveSelectNone config to multi-select

ces131:
Adds a config to prevent users from selecting all of the survey options and a config to remove the "select none/de-select all" option in multiselect. Using `survey.WithRemoveSelectAll()` will disable the ability for the right arrow key to select all of the options. It will also remove the instructions from the multi-select prompt.
Similarly, `survey.WithRemoveSelectNone()` will disable not allow a user to select none/de-select all options and the instruction won't show up in the prompt.

Aurora Innovation has a patch with this work in their repo, and I figured I would make the change upstream. I filed a ticket quite awhile ago and finally got around to adding the change: https://github.com/AlecAivazis/survey/issues/393

### Test Plan

Ran all of the tests. Added a new test that verifies the correct prompt and checks that using the right arrow key doesn't select all the options and using the left arrow key doesn't de-select all the options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -357,7 +357,7 @@ All of the prompts have a `Help` field which can be defined to provide more info
 }
 ```
 
-## Removing the "Select All" option
+## Removing the "Select All" and "Select None" options
 
 By default, users can select all of the multi-select options using the right arrow key. To prevent users from being able to do this (and remove the `<right> to all` message from the prompt), use the option `WithRemoveSelectAll`:
 
@@ -372,6 +372,21 @@ prompt := &survey.Input{
 }
 
 survey.AskOne(prompt, &number, survey.WithRemoveSelectAll())
+```
+
+Also by default, users can use the left arrow key to unselect all of the options. To prevent users from being able to do this (and remove the `<left> to none` message from the prompt), use the option `WithRemoveSelectNone`:
+
+```golang
+import (
+    "github.com/AlecAivazis/survey/v2"
+)
+
+number := ""
+prompt := &survey.Input{
+    Message: "This question has the select all option removed",
+}
+
+survey.AskOne(prompt, &number, survey.WithRemoveSelectNone())
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -357,6 +357,24 @@ All of the prompts have a `Help` field which can be defined to provide more info
 }
 ```
 
+## Removing the "Select All" option
+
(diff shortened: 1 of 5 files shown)


AlecAivazis:
Thanks for putting this together @ces131! Should removing the select all also remove the select none behavior? Maybe we want a second option while we're here?

ces131:
> Thanks for putting this together @ces131! Should removing the select all also remove the select none behavior? Maybe we want a second option while we're here?

If you'd like, I can add an additional option `RemoveSelectNone` while I'm here - for our use case I think we'd like to keep the "none" option even with the SelectAll removed. Sometimes our mulit-select question is "preloaded"/has a few options pre-selected so a quick way for the user to remove all of those is a nice to have

AlecAivazis:
Yea i think we should just for completeness. thanks!

survey.go
```diff
@@ -175,6 +179,22 @@ func WithKeepFilter(KeepFilter bool) AskOpt {
 	}
 }
 
+// WithRemoveSelectAll remove the select all option in Multiselect
+func WithRemoveSelectAll(RemoveSelectAll bool) AskOpt {
```
AlecAivazis, line 183: I think given that the name of the option is "Remove..." it's a bit confusing to accept a parameter. Does saying Remove(false) turn it on? I think it's probably less confusing if we don't accept any arguments.

README.md
```diff
@@ -357,6 +357,39 @@ All of the prompts have a `Help` field which can be defined to provide more info
...
+import (
+    "github.com/AlecAivazis/survey/v2"
+)
+
+number := ""
+prompt := &survey.Input{
+    Message: "This question has the select all option removed",
+}
+
+survey.AskOne(prompt, &number, survey.WithRemoveSelectAll())
```
AlecAivazis, line 374: We need this to be kept up to date with how things shake out in other comment.

AlecAivazis (changes requested):
Sorry this review is coming in two pieces but I've been struggling to find the time. Anyway, we're almost there - one small tweak and this is good to go

survey.go
ces131: Ahh, yes I totally agree with this (I guess that shows through the docs being mixmatched) I think I was following the method above `WithKeepFilter(KeepFilter bool)`.

Updated these both to no longer have a bool as an arg here

AlecAivazis (approved):
Looks good! thanks for updating the docs (and the ping 😅 )

ces131:
> Looks good! thanks for updating the docs (and the ping 😅 )

No problem at all! Thanks for the review 😄

PR #442, opened 2022-08-13
Fix bugs when deleting characters when prompt is at the bottom of the terminal

### LINKED ISSUE #437: Last line in prompt becomes invisible when a character is typed and then deleted on a previous line

MacOS 12.4; alacritty (but bug is reproducible in any terminal)

This is a bug found in GitHub's CLI that uses the `survey` library.
See [this issue in GitHub's CLI](https://github.com/cli/cli/issues/6015#issuecomment-1203709949).

1. Open new terminal window in a git repo
2. Press enter many times until the terminal scrolls down
3. Run `gh issue create`
4. Type text in the `Title` prompt until it wraps to a new line at least once
5. Move the cursor to a previous line before the line wrap
6. Type a character and delete it

Capture d’écran 2022-07-31 à 17 03 11

Capture d’écran 2022-07-31 à 17 03 24

Capture d’écran 2022-07-31 à 17 03 39

Text on the last line

**What did you see instead?**

No text on the last line

pokir:
This fixes #437 by checking if the cursor moved down a line before erasing the line.

When the terminal scrolls down after the text reaches the bottom of the terminal, no next line exists.
So trying to move the cursor down one line just makes it go to the beginning of the same line.
Erasing the line then erases the last line of text, making it invisible.

### DIFF
--- a/terminal/runereader.go
+++ b/terminal/runereader.go
@@ -179,18 +179,17 @@ func (rr *RuneReader) ReadLineWithDefault(mask rune, d []rune, onRunes ...OnRune
 					EraseLine(rr.stdio.Out, ERASE_LINE_END)
 
 					// Erase symbols which are left over from older print on the next line
-					if cursorCurrent.Y < terminalSize.Y {
-						// store the cursor position to compare it later
-						lastCursorLocation, _ := cursor.Location(rr.Buffer())
 
-						// this moves the cursor down or stays on the same line if no next line exists
-						cursor.NextLine(1)
+					// store the cursor position to compare it later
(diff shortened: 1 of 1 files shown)


pokir:
Also fixed a similar bug, where if there was only one character on the last line of the terminal and a character was deleted on a previous line, it wouldn't get erased.

AlecAivazis:
👋 Thanks for submitting this! Please bear with us while we catch up on a lot of PRs that are piling up. We appreciate your patience

AlecAivazis:
I'm going to close this since its gone a bit stale and I can't reproduce it.

PR #450, opened 2022-09-14
Show transformed answer after submitting input

### LINKED ISSUE #449: Transformed answer not displayed after submitting input

- macOS 12.5.1 (Monterey)
- Terminal.app

```go
package main

import (
	"fmt"
	"strings"

)

func main() {
	qs := []*survey.Question{
		{
			Prompt: &survey.Input{
				Message: "Name",
			},
			Transform: survey.TransformString(strings.TrimSpace),
		},
	}
	name := ""
	_ = survey.Ask(qs, &name)

	fmt.Println("Hello,", name)
}
```

Ran the above and submitted: 

```
            Skip
```

```
[~/tmp] $ go run main.go 
? Name Skip
Hello, Skip
```

**What did you see instead?**

```
[~/tmp] $ go run main.go 
? Name             Skip
Hello, Skip
```

The transformed answer [is passed to the prompt on `Cleanup()`](https://github.com/AlecAivazis/survey/blob/bf4b865e6c271af51ee152a74e41a037d1f8b013/survey.go#L365-L377), but appears to be ignored in favor of the input's existing answer:

https://github.com/AlecAivazis/survey/blob/bf4b865e6c271af51ee152a74e41a037d1f8b013/input.go#L211

Changing that to `ans := val.(string)` seems to fix the issue. I'll try writing a test case and submit a PR.

twelvelabs:
Fixes #449.

I tried to use the existing `go-expect` tooling to test this, but had a hard time figuring out how to do so :grimacing:. No amount of fiddling w/ `Console` (or the `Expect*` methods) seemed to be able to get at the output written by `Prompt.Confirm()`. If there's a better way to do this, please let me know and I'll update.

### DIFF
--- a/input.go
+++ b/input.go
@@ -207,20 +207,13 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
 }
 
 func (i *Input) Cleanup(config *PromptConfig, val interface{}) error {
-	// use the default answer when cleaning up the prompt if necessary
-	ans := val.(string)
-	if ans == "" && i.Default != "" {
-		ans = i.Default
-	}
-
-	// render the cleanup
 	return i.Render(
 		InputQuestionTemplate,
 		InputTemplateData{
 			Input:      *i,
 			ShowAnswer: true,
 			Config:     config,
-			Answer:     ans,
+			Answer:     val.(string),
 		},
 	)
 }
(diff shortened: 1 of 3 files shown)


mislav (approved):
Good catch!

I've pushed some changes to simplify. Let me know what you think

twelvelabs:
Ah, it seems I misunderstood the failure message when I originally tried using `ExpectString()`. That was what I was going for - thanks!

LGTM 👍

Issue #368, opened 2021-08-12
Duplicated prompt after enter input


yomono:
Hi, I'm getting this problem:

![image](https://user-images.githubusercontent.com/31445936/129165091-8aae8d67-90ec-452a-a452-ec6d85b8aa2f.png)

This is happening with the "Input" survey. Not happening on select, for example.
Note that the image is showing the "cursor" example from this repo.

Is this a bug or am I doing something wrong?

Thanks in advance

raklaptudirm:
Any updates?

QuantumAI:
Issue seems to be present since v2.2.14, not in v2.2.13

PR #458, opened 2022-10-12
fix issue with double output on windows

### LINKED ISSUE #406: Randomly duplicated output for Confirm prompt with a non default answer
Ubuntu 21.10 (this also runs on my CI which is painful)
Mac OSX 

When using unit tests I sometimes see this behaviour on Ubuntu, I think it is related to https://github.com/AlecAivazis/survey/pull/404

```
?Do you like pizza? Yes
```
**What did you see instead?**
```
?Do you like pizza (y/N)? y
?Do you like pizza? Yes
```

miniscruff:
I was able to fix the double output issues on windows by removing this line, did a manual test on Ubuntu as well. My guess is this is not the best solution so let me know.

Fixes #406 and #368

### DIFF
--- a/input.go
+++ b/input.go
@@ -200,8 +200,6 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
 
 	lineStr := i.answer
 
-	i.AppendRenderedText(lineStr)
-
 	// we're done
 	return lineStr, err
 }


input.go
```diff
@@ -200,8 +200,6 @@ func (i *Input) Prompt(config *PromptConfig) (interface{}, error) {
 
 	lineStr := i.answer
 
-	i.AppendRenderedText(lineStr)
```
mislav, line 203: This code gets called on all platforms and its implementation doesn't differ between platforms. Do you know how it could have caused problems only on Windows but not elsewhere?

mislav (changes requested):
Hi, sorry for the late review and thanks for helping fix this. I think what this PR would need is some kind of regression test that fails without this fix being applied and will fail in the future if this regresses.

input.go
miniscruff: I am not entirely sure what happens on windows to cause this problem, I can try and dig deeper but my only guess right now has to do with the carriage return, as it seems to render the same text twice and this was to remove the second one, which is probably totally wrong. I did do testing on Ubuntu as well and found there was no difference in outputs of my test between before and after my change. Obviously I could be missing some manual test cases so if there are more I can do let me know.

mislav:
> I was trying to add one of these but the functional tests do not run on windows

Ah, you are right. I'll investigate this further with your changes as a starting point. Thank you!

miniscruff:
> > I was trying to add one of these but the functional tests do not run on windows
> 
> Ah, you are right. I'll investigate this further with your changes as a starting point. Thank you!

No problem, let me know if I can be of any help.

mislav:
@miniscruff Could you please try https://github.com/go-survey/survey/pull/474?

miniscruff:
> @miniscruff Could you please try #474?

That seems to have fixed the problem for me in windows terminal yes.

miniscruff:
@mislav your PR will take care of the issue so we can probably close this PR correct?

mislav:
Yes. Thanks for the reminder!

PR #461, opened 2022-10-21
feat: add regexp validator

lzakharov:


mislav (commented):
Thanks! The code looks fine to me but I'm not sure if this is a feature we'd want in Survey core. It's trivial to add something like this to one's own application, with the added benefit that the user can control the language in the error messages that way.

@AlecAivazis How do you feel about this feature?

AlecAivazis:
Yea I agree with you @mislav. For me, the biggest reason is the error message. The user of the cli shouldn't really know that the tool is using a regexp to validate their input so having this message leak that out feels a little awkward / too low-level

lzakharov:
Thanks for the comments, got you!:)

PR #462, opened 2022-10-30
Remove ANSI formatting before measuring string width

e-zim:
### Summary

Hello! This PR hopes to address an issue in the select prompt where formatted lines may delete the past output on terminals of certain sizes.

### The problem

When [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code) are used in the options of a prompt, these lines are measured wider than what is visibly printed to the terminal. This is caused by the `StringWidth` counting these escaped characters as a part of the actual string, and results in previous output being deleted (if the measured line count is greater than the actual line count of an option).

The following can demonstrate this on terminals that are only slightly wider than the option's visible length:

```go
prompt := &survey.Select{
  Message: "Select an option:",
  Options: []string{
    "\033[35m..........................................................",
    "\033[36m..........................................................",
  },
}

survey.AskOne(prompt, "")
```

See this in action:

https://user-images.githubusercontent.com/18134219/198906355-71ca8926-61e6-4af0-8c44-9adf03a8604a.mov

### Proposed solution

The [`github.com/acarl005/stripansi` library](https://github.com/acarl005/stripansi) offers a nice regex pattern to remove ANSI codes, and when used in the `StringWidth` method can remove these codes before the string's width is measured. 

Additionally, a check for non-printable runes was added to the `runeWidth` method so that only printable characters are measured. [`unicode.IsPrint`](https://pkg.go.dev/unicode#IsPrint) was chosen over [`unicode.IsGraphic`](https://pkg.go.dev/unicode#IsGraphic) to conform with the definitions of Go, but I could see a reason for preferring `IsGraphic`!

With the changes of this PR applied, the following result is achieved:

https://user-images.githubusercontent.com/18134219/198907296-7fcfa73a-4f78-4a20-990a-f203f03202b7.mov

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -1,7 +1,5 @@
 github.com/Netflix/go-expect v0.0.0-20220104043353-73e0943537d2 h1:+vx7roKuyA63nhn5WAunQHLTznkw5W8b1Xc0dNjp83s=
 github.com/Netflix/go-expect v0.0.0-20220104043353-73e0943537d2/go.mod h1:HBCaDeC1lPdgDeDbhX8XFpy1jqjK0IBG8W5K+xYqA0w=
-github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d h1:licZJFw2RwpHMqeKTCYkitsPqHNxTmd4SNR5r94FGM8=
-github.com/acarl005/stripansi v0.0.0-20180116102854-5a71ef0e047d/go.mod h1:asat636LX7Bqt5lYEZ27JNDcqxfjdBQuJ/MM4CN/Lzo=
 github.com/creack/pty v1.1.17 h1:QeVUsEDNrLBW4tMgZHvxy18sKtr6VI492kBhUfhDJNI=
 github.com/creack/pty v1.1.17/go.mod h1:MOBLtS5ELjhRRrroQr9kyvTxUAFNvYEK993ew/Vr4O4=
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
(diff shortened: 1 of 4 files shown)


terminal/runereader.go
```diff
@@ -377,17 +378,24 @@ func (rr *RuneReader) ReadLineWithDefault(mask rune, d []rune, onRunes ...OnRune
...
+		return 0
+	}
 	return 1
 }
 
+// StringWidth returns the visible width of a string when printed to the terminal
 func StringWidth(str string) int {
 	w := 0
-	rs := []rune(str)
+	cleanedStr := stripansi.Strip(str)
```
mislav, line 397: I think it's an overkill to pull in a dependency just to get access to a single regular expression. However, since we're looping to measure the printable string width here, I would be more in favor of an approach like this one: https://github.com/muesli/reflow/blob/v0.3.0/ansi/buffer.go#L20-L40

I'm also not suggesting to import `github.com/muesli/reflow/ansi` here, but we could inline the approach of ignoring ANSI escape codes while iterating over a string. What do you think?

mislav (changes requested):
Thanks for proposing to fix this!

terminal/runereader.go
e-zim: I similarly agree that a whole import for this one line might be too much.. I also really like the iterative approach as it's slightly more readable than the regex too! I'll make these changes in the next day or two!

go.sum
```diff
@@ -21,7 +21,6 @@ github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+
 github.com/stretchr/testify v1.6.1 h1:hDPOHmpOpP40lSULcqw7IrRb/u7w6RpDC9399XyoNd0=
 github.com/stretchr/testify v1.6.1/go.mod h1:6Fq8oRcR53rry900zMqJjRRixrwX3KX962/h/Wwjteg=
 golang.org/x/sys v0.0.0-20190222072716-a9d3bda3a223/go.mod h1:STP8DvDyc/dI5b8T5hshtkjS+E42TnysNCUPdjciGhY=
-golang.org/x/sys v0.0.0-20201119102817-f84b799fce68 h1:nxC68pudNYkKU6jWhgrqdreuFiOQWj1Fs7T3VrH4Pjw=
```
e-zim, line 24: This is a result of running `go mod tidy` after removing the `stripansi` package. If keeping this line feels safer, that's good with me!

terminal/runereader.go
```diff
@@ -377,19 +377,42 @@ func (rr *RuneReader) ReadLineWithDefault(mask rune, d []rune, onRunes ...OnRune
...
 }
 
+// isAnsiMarker returns if a rune denotes the start of an ANSI sequence
+func isAnsiMarker(r rune) bool {
+	return r == '\x1B'
+}
+
+// isAnsiTerminator returns if a rune denotes the end of an ANSI sequence
+func isAnsiTerminator(r rune) bool {
+	return (r >= 0x40 && r <= 0x5a) || (r == 0x5e) || (r >= 0x60 && r <= 0x7e)
```
e-zim, line 400: These terminators were collected from various lists and other examples to match [the XTerm Control Sequences](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_).

mislav (approved):

mislav:
Thank you!

mwbrooks:
@mislav Some wonderful improvements and bug fixes have landed since [September's v2.3.6 release](https://github.com/go-survey/survey/releases/tag/v2.3.6). When you have a moment in the coming months, we'd greatly appreciate a fresh release 🙏🏻 ☺️

PR #469, opened 2022-11-22
Doc update.

p000ic:
Updated imports in the sample code.


mislav (commented):
Thanks, but note that it's not possible yet to fetch this package using the new repo URL because our package name is still the old name in our `go.mod`.

We will resolve this together with the fix for https://github.com/go-survey/survey/issues/465

PR #474, opened 2022-11-24
Fix redrawing prompts on Windows

mislav:
Fixes redrawing prompts on Windows so that:
1. The line before the 1st survey prompt isn't accidentally deleted,
2. The survey prompt is properly redrawn after recording an answer.

For help with debugging Survey, this also adds logging support by setting SURVEY_LOG_FILE environment variable.

Fixes https://github.com/go-survey/survey/issues/368
Closes https://github.com/go-survey/survey/pull/458

TODO:
- [ ] test in more different terminal combos
- [ ] test cursor handling when a prompt or an answer wraps due to viewport width


andyfeller:
@mislav : thank you for working towards the root of this issue for Windows users ✨ 

**What might you share about this work for someone else to pick it up and carry it completion?**

PR #475, opened 2022-12-03
Output plain template when NO_COLOR environment variable is set

### LINKED ISSUE #447: Support NO_COLOR Environment Variable
Survey should respect a `NO_COLOR` environment variable for reasons articulated at https://no-color.org/ and https://clig.dev/#output. Currently, Survey emits text with colors no matter what value is specified for this environment variable.

e-zim:
### Summary

This PR checks for a non-empty `NO_COLOR` environment variable and displays templates without color if one is present, as specified by the references in #447.

Fixes #447

### Preview

https://user-images.githubusercontent.com/18134219/205460922-8b6a854b-e4e6-4daf-93b8-3e8ed63ca6be.mov

### Reviewers

The following steps can be used to inspect these changes:

1. Checkout this branch
2. In a project using `survey`, append the following to your `go.mod`, pointing to your local `survey` source:

```
replace github.com/AlecAivazis/survey/v2 v2.3.6 => ../../go-survey/survey
```

3. Run your project with color support by ensuring `NO_COLOR` is not set:

```sh
$ unset NO_COLOR
$ ./your-survey-project
```

4. Run your project with colors removed from `survey` outputs:

```sh
$ export NO_COLOR=true
$ ./your-survey-project
```

### Notes

- No modifications are made to user-passed prompt values such as `Message`, `Options`, `Help`, and so forth. Removing color from these values is instead left to the caller of this package.

### DIFF
--- a/core/template.go
+++ b/core/template.go
@@ -24,6 +24,17 @@ var TemplateFuncsNoColor = map[string]interface{}{
 	},
 }
 
+// envColorDisabled returns if output colors are forbid by environment variables
+func envColorDisabled() bool {
+	return os.Getenv("NO_COLOR") != "" || os.Getenv("CLICOLOR") == "0"
+}
+
+// envColorForced returns if output colors are forced from environment variables
+func envColorForced() bool {
+	val, ok := os.LookupEnv("CLICOLOR_FORCE")
+	return ok && val != "0"
+}
+
 // RunTemplate returns two formatted strings given a template and
 // the data it requires. The first string returned is generated for
 // user-facing output and may or may not contain ANSI escape codes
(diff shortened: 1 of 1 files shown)


core/template.go
```diff
@@ -23,6 +24,17 @@ var TemplateFuncsNoColor = map[string]interface{}{
...
 
+// envColorDisabled returns if output colors are forbid by environment variables
+func envColorDisabled() bool {
+	return os.Getenv("NO_COLOR") != "" || os.Getenv("CLICOLOR") == "0"
+}
+
+// envColorForced returns if output colors are forced from environment variables
+func envColorForced() bool {
+	val, ok := os.LookupEnv("CLICOLOR_FORCE")
+	return ok && val != "0"
```
e-zim, line 35: Happy to change this to match the reference if that's preferred, but wanted to share a way to check if the environment variable is present then check for a `!= 0` value instead of checking that the variable isn't the empty value.

mislav (approved):
Thank you! Looks great.

@AlecAivazis How do you feel about having Survey respect environment variables by default?

AlecAivazis:
Seems totally fine to me 👍

Issue #452, opened 2022-09-19
Suggestion problem


nemati21:
Hi,
I have an issue with the suggestion when I press enter button to select an item, the previous question was removed from CLI. 
I don't know if it's a bug or my mistake.
Please help me.

nhatthm:
I have the same problem, tested by compiling with all the versions back til 2.2.13 which doesn't have the issue. [This is the only PR](https://github.com/AlecAivazis/survey/pull/361) in 2.2.14

nhatthm:
If I remove this line

https://github.com/AlecAivazis/survey/blob/a98a037e3f0b51b8a992bfdd4c50debb70a53553/input.go#L186

The answer is correctly printed after selecting a suggestion.

However, the random answer (without selecting a suggestion) is not, it's printed one line after the prompt. For example

```
? Topic [tab for suggestions] asd
? Topic asd
```

nemati21:
I customized the InputQuestionTemplate and removed "enter to select" from the shown message that the user uses the right arrow to select an item but it isn't a good solution. it must be fixed

nhatthm:
Code

```go
package main

import (
	"fmt"
	"os"
	"strings"

	"github.com/AlecAivazis/survey/v2"
)

func main() {
	var answer string

	suggests := []string{"Amelia", "Ava", "Elijah", "Emma", "Liam", "Mateo", "Noah", "Oliver", "Olivia", "Sophia"}

	err := survey.AskOne(&survey.Input{
		Message: "What is your name?",
		Suggest: func(toComplete string) []string {
			if len(toComplete) == 0 {
				return suggests
			}

			var matches []string

			for _, s := range suggests {
				if strings.Contains(s, toComplete) {
					matches = append(matches, s)
				}
			}

			return matches
		},
	}, &answer)

	if err != nil {
		fmt.Fprintln(os.Stderr, err.Error())
		os.Exit(1)
	}

	// print the answers
	fmt.Printf("Hi %s.\n", answer)
}
```

Env:
- macOS 12.6
- zsh 5.9
- iTerm2 3.4.16

Stable version `2.3.6`

https://user-images.githubusercontent.com/1154587/192954112-80f27f6f-861e-4e89-996b-f00e4a0a3d5b.mov

Last working version `2.2.13`

https://user-images.githubusercontent.com/1154587/192954747-9c09858d-071e-41a3-850a-b23b9a2c3e4e.mov

nemati21:
`package main

import (
[...]

Issue #436, opened 2022-07-18
Terminal consistently flickers when scrolling through paginated list of options


aq17:
**What operating system and terminal are you using?**
MacOS+default terminal, also saw same issue on Windows+PowerShell

**An example that showcases the bug.**
(short clip to fit 10mb limit, but continues as you scroll down the list and paginate)
https://user-images.githubusercontent.com/54565933/179568187-b4d8e098-a23b-4484-b988-ca143dda607a.mov

**What did you expect to see?**
No flickering when scrolling through paginated list

**What did you see instead?**
Flickering when scrolling through paginated list

jquick:
I have this issue also

ekini:
Same behaviour on Linux (Wayland) in "foot" terminal.

mislav:
I also experience this and wanted to find ways to mitigate. This is the redrawing logic: https://github.com/AlecAivazis/survey/blob/e47352f914346a910cc7e1ca9f65a7ac0674449a/renderer.go#L135-L145

Perhaps there is a way to redraw the last N lines in a terminal in a way that doesn't cause flickering. Ideas welcome!

dnkl:
> Perhaps there is a way to redraw the last N lines in a terminal in a way that doesn't cause flickering. Ideas welcome!

There are application synchronized updates: https://gitlab.freedesktop.org/terminal-wg/specifications/-/merge_requests/2

TL;DR application emits one escape at the beginning of a "frame" (update) and another one when done.

Note that there are two variants; the DCS initially suggested in the MR, which was the first one implemented by terminals. Later, it was generally agreed that a "private mode" was better, and the `2026` private mode was added.

Another option is writing the erase commands into a buffer, and doing a single actual terminal write.

Issue #336, opened 2021-02-26
Show multiline's answer in new line


AielloChan:
Now, it looks wired:

![image](https://user-images.githubusercontent.com/7900765/109301804-653cfd00-7873-11eb-81cc-2b36aa7d5e54.png)

How about showing the answer in a new line, just like below:

![image](https://user-images.githubusercontent.com/7900765/109302731-a681dc80-7874-11eb-85f2-457c1e146fff.png)

AielloChan:
And it just needs and a `\n` in the template.

![image](https://user-images.githubusercontent.com/7900765/109304197-dd58f200-7876-11eb-95ee-9257b39b7fe8.png)


https://github.com/AielloChan/survey/commit/44148ffac0f0a13e6031fb046338aff69e01a889

AielloChan:
So, any idea?

AlecAivazis:
Thanks for the ping @AielloChan! It's been hard to wrangle all of the different notifications I get so I appreciate your patience.

At the moment, a change like this would be considered a break in the API. `survey` is used by many large companies who run integration tests and a change like this would inevitably make those fail.

I know its not the answer you or the others want to hear but I dont think we can change this at the moment. I will leave the issue open and tag it for future consideration though!

AielloChan:
> Thanks for the ping @AielloChan! It's been hard to wrangle all of the different notifications I get so I appreciate your patience.
> 
> At the moment, a change like this would be considered a break in the API. `survey` is used by many large companies who run integration tests and a change like this would inevitably make those fail.
> 
> I know its not the answer you or the others want to hear but I dont think we can change this at the moment. I will leave the issue open and tag it for future consideration though!

Okay, this would be the best choice for now. 😄

graugans:
I agree with @AielloChan the current behavior bugged me  as-well. I am just staring with Go so am I far from being one which could provide any implementation advice nor I fully understand how survey is internally working.

Given the templating engine of Go and an extension of [PromptConfig](https://github.com/AlecAivazis/survey/blob/099a96894e7a1941cca62b6a9f1a06dadf8664ae/survey.go#L113) it should be possible to flag this feature. So no integration test on larger companies will break

silvanbrummen:
I'm running into this styling issue myself as well currently. The solution proposed by @graugans seems like a good option to me. What do you think @AlecAivazis?

CastawayEGR:
+1

sheldonhull:
[...]

PR #476, opened 2022-12-04
Clear to end of screen before redrawing prompt

e-zim:
### Summary

This PR adds a `terminal.EraseScreen` function to support the "Erase in Display" ANSI control sequence. This function replaces calls to `terminal.ClearLine` in `resetPrompt` in order to clear previous prompt output entirely and atomically, instead of clearing each line in an iterative manner.

Updating `resetPrompt` to clear to the end of the screen seems to significantly reduce the "flickering" effect noted in #436 since fewer updates to the terminal screen are made between each render.

Additionally, the `MultiLine` prompt was updated as a result of changes to the `resetPrompt` logic. This prompt now preserves leading and trailing spaces in answers and replaces the input template with an answered template after submission. The template was also updated to begin inputs on a newline, which happens to be a request of #336.

#### Before

https://user-images.githubusercontent.com/18134219/205465447-cda292c1-bb7b-4ef3-af26-aa450b00791b.mp4

#### After

https://user-images.githubusercontent.com/18134219/205465499-ed0040e4-80ec-49e8-b0df-9223258a1b10.mp4

### Reviewers

The following steps can be used to inspect these changes:

1. Checkout this branch
2. In a project using `survey` and `survey.Select`, `survey.MultiSelect`, or `survey.MultiLine`, append the following to your `go.mod`, pointing to your local `survey` source:
```
replace github.com/AlecAivazis/survey/v2 v2.3.6 => ../../go-survey/survey
```

3. Run your project and rapidly page through select options in an attempt to reproduce the flickering effect.

### Notes

- Flickering is not entirely removed since there is brief moment between clearing the past prompt and outputting the updated one where the erased screen might be shown. This seems to be a fairly rare occurrence though, only happening a handful of times in my testing. The "synchronized update" suggestion from @dnkl would likely prevent this, but I had some difficulties in implementing this.

- This has not been tested with a Windows terminal, however I presume this escape code is properly handled as "[virtual terminal sequences are recommended](https://learn.microsoft.com/en-us/windows/console/classic-vs-vt#recommendations)" when developing for Windows.
  - These changes are also likely to impact the updates in #474, as both PRs are modifying the logic of `resetPrompt`.

- These changes do not seem to impact the current behavior shown in #452.

### DIFF
--- a/multiline.go
+++ b/multiline.go
@@ -86,12 +86,12 @@ func (i *Multiline) Prompt(config *PromptConfig) (interface{}, error) {
 		multiline = append(multiline, string(line))
 	}
 
+	// remove the trailing newline
+	multiline = multiline[:len(multiline)-1]
 	val := strings.Join(multiline, "\n")
-	val = strings.TrimSpace(val)
 
-	// if the line is empty
+	// use the default value if the line is empty
 	if len(val) == 0 {
-		// use the default value
 		return i.Default, err
 	}
 
--- a/multiline.go
+++ b/multiline.go
@@ -2,8 +2,6 @@ package survey
 
 import (
 	"strings"
-
-	"github.com/AlecAivazis/survey/v2/terminal"
 )
 
 type Multiline struct {
@@ -28,12 +26,12 @@ var MultilineQuestionTemplate = `
 {{- color .Config.Icons.Question.Format }}{{ .Config.Icons.Question.Text }} {{color "reset"}}
 {{- color "default+hb"}}{{ .Message }} {{color "reset"}}
 {{- if .ShowAnswer}}
-  {{- "\n"}}{{color "cyan"}}{{.Answer}}{{color "reset"}}
-  {{- if .Answer }}{{ "\n" }}{{ end }}
+  {{- if .Answer}}{{"\n"}}{{color "cyan"}}{{.Answer}}{{color "reset"}}{{end}}
 {{- else }}
   {{- if .Default}}{{color "white"}}({{.Default}}) {{color "reset"}}{{end}}
   {{- color "cyan"}}[Enter 2 empty lines to finish]{{color "reset"}}
-{{- end}}`
+{{- end}}
+`
 
 func (i *Multiline) Prompt(config *PromptConfig) (interface{}, error) {
 	// render the template
@@ -70,13 +68,6 @@ func (i *Multiline) Prompt(config *PromptConfig) (interface{}, error) {
 
 		if string(line) == "" {
 			if emptyOnce {
-				numLines := len(multiline) + 2
-				cursor.PreviousLine(numLines)
-				for j := 0; j < numLines; j++ {
-					terminal.EraseLine(i.Stdio().Out, terminal.ERASE_LINE_ALL)
-					cursor.NextLine(1)
-				}
-				cursor.PreviousLine(numLines)
 				break
 			}
 			emptyOnce = true
@@ -85,6 +76,15 @@ func (i *Multiline) Prompt(config *PromptConfig) (interface{}, error) {
 		}
 		multiline = append(multiline, string(line))
 	}
+
+	// position the cursor on last line of input
+	cursorPadding := 3
+
+	// ignore the empty line in an empty answer
+	if len(multiline) == 1 && multiline[0] == "" {
+		cursorPadding = 2
+	}
+	cursor.PreviousLine(cursorPadding)
 
 	// remove the trailing newline
 	multiline = multiline[:len(multiline)-1]
--- a/terminal/cursor.go
+++ b/terminal/cursor.go
@@ -47,18 +47,24 @@ func (c *Cursor) Back(n int) error {
 
 // NextLine moves cursor to beginning of the line n lines down.
 func (c *Cursor) NextLine(n int) error {
-	if err := c.Down(1); err != nil {
+	if err := c.HorizontalAbsolute(0); err != nil {
 		return err
 	}
-	return c.HorizontalAbsolute(0)
+	if n == 0 {
+		return nil
+	}
+	return c.Down(n)
 }
 
 // PreviousLine moves cursor to beginning of the line n lines up.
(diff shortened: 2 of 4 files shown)


terminal/cursor.go
```diff
@@ -47,18 +47,24 @@ func (c *Cursor) Back(n int) error {
 
 // NextLine moves cursor to beginning of the line n lines down.
 func (c *Cursor) NextLine(n int) error {
-	if err := c.Down(1); err != nil {
+	if err := c.HorizontalAbsolute(0); err != nil {
```
mislav, line 50: Why first move horizontal and then try to move between lines?

terminal/cursor.go
```diff
@@ -47,18 +47,24 @@ func (c *Cursor) Back(n int) error {
...
 		return err
 	}
-	return c.HorizontalAbsolute(0)
+	if n == 0 {
+		return nil
+	}
+	return c.Down(n)
```
mislav, line 56: Nice catch that this didn't use to forward the `n` parameter and just moved down 1 line each time 👍

multiline.go
```diff
@@ -86,12 +77,21 @@ func (i *Multiline) Prompt(config *PromptConfig) (interface{}, error) {
...
 	}
 
+	// position the cursor on last line of input
+	cursorPadding := 3
+
+	// ignore the empty line in an empty answer
+	if len(multiline) == 1 && multiline[0] == "" {
+		cursorPadding = 2
+	}
+	cursor.PreviousLine(cursorPadding)
```
mislav, line 87: Let's make this simpler to read at the cost of tiny repetition:
```suggestion
	if len(multiline) == 1 && multiline[0] == "" {
		// ignore the empty line in an empty answer
		cursor.PreviousLine(2)
	} else {
		// position the cursor on last line of input
		cursor.PreviousLine(3)
	}
```
I'm curious about where these constants `2` and `3` come from. From the old code, it looks like the number of lines to clear was calculated by taking the length of `multiline` and adding 2 to that (I guess to account for two blank lines). Now, it either deletes 2 or 3 lines up and not more than that. Is the size of `multiline` taken into account somewhere when redrawing?

terminal/display.go
```diff
@@ -1,9 +1,25 @@
...
 )
+
+const (
+	ERASE_SCREEN_END EraseScreenMode = iota
+	ERASE_SCREEN_START
+	ERASE_SCREEN_ALL
+)
+
+func EraseScreen(out FileWriter, mode EraseScreenMode) error {
+	_, err := fmt.Fprintf(out, "\x1b[%dJ", mode)
```
mislav, line 23: on Windows we can't just assume that printing `\x1b[%dJ` to the terminal will just work. I'm pretty sure that out-of-the-box, that won't work in both cmd.exe and PowerShell.exe terminal, and seeing how Survey right now works in both, we would have to come up with a solution to clear to the end of the terminal that would work in all terminals that Survey currently supports.

mislav (changes requested):
Thanks for the deep dive into this! I love to see improvements around making fewer individual erase operations in favor of just one.

terminal/cursor.go
e-zim: There seems to be weird behavior when going `Down(0)` or `Up(0)`, where the cursor moves down or up `1` instead of remaining on the same line. I'm not sure if this is [expected ANSI behavior](https://en.wikipedia.org/wiki/ANSI_escape_code#CSI_(Control_Sequence_Introducer)_sequences) for `\x1b[0A`, but this change was made to guard against this case.

It might be more appropriate to move this check for `n == 0` to the `Down` and `Up` functions instead? I wasn't sure if this current behavior was relied on by other functions and felt that this change was safer.

terminal/display.go
e-zim: Completely agree! I'd be interested in knowing if a modified version of [this example](https://learn.microsoft.com/en-us/windows/console/clearing-the-screen#example-3) could be used to clear only from the current cursor position to the end of the screen? I have hope that the `FillConsoleOutputCharacter` call is buffered and would fill the remaining screen with blanks in a single action, which should mean that option flickering is removed from select prompts on Windows, but I am unfamiliar with these calls and am open to any suggestions!

terminal/display_windows.go
```diff
@@ -29,3 +29,22 @@ func EraseLine(out FileWriter, mode EraseLineMode) error {
...
+	if _, _, err := procGetConsoleScreenBufferInfo.Call(uintptr(handle), uintptr(unsafe.Pointer(&csbi))); normalizeError(err) != nil {
+		return err
+	}
+
+	var w uint32
+	cursor := csbi.cursorPosition
+
+	lineCount := csbi.window.bottom - csbi.cursorPosition.Y
+	termWidth := csbi.size.X
+	screenSize := lineCount * termWidth
```
e-zim, line 46: The Windows version of this `EraseScreen` function clears the lines from the cursor (at [the beginning of the prompt](https://github.com/e-zim/survey/blob/d29d6fa9a7a31b354b33c29b6128b504fd3c6af6/renderer.go#L135)) to the end of the displayed window (which is the end of the outputted prompt). For reference, [the `consoleScreenBufferInfo` structure](https://learn.microsoft.com/en-us/windows/console/console-screen-buffer-info-str) contains terminal dimensions and coordinates of the window and cursor.

This seems to match the behavior of the POSIX `EraseScreen` for resetting the prompt, but doesn't account for the `EraseScreenMode`. Not sure if supporting the other modes (clearing to the beginning of the screen and the entire screen) is necessary for how this function is being used, but I'd be in favor of forgoing them to keep this implementation more simple. Definitely open to thoughts here!

e-zim (commented):
Added support for Windows terminals and tested in both cmd.exe and Powershell!

Unfortunately, it seems that there's still the occasional flickering effect with this redraw logic, but it doesn't seem to be any better or worse than currently. Just a different way of achieving the same effect. This might be more pronounced on my older machine though!

PR #482, opened 2023-04-03
Allow hiding password length

szh:
Addresses [comments](https://github.com/go-survey/survey/issues/386#issuecomment-1494792529) to #386

This PR allows you to set the `HideCharacter` to a space which will be interpreted by the RuneReader to mean that the user input should not be echoed and that it shouldn't print a "mask" either.

### DIFF
--- a/survey.go
+++ b/survey.go
@@ -256,7 +256,8 @@ func WithShowCursor(ShowCursor bool) AskOpt {
 	}
 }
 
-// WithHideCharacter sets the default character shown instead of the password for password inputs
+// WithHideCharacter sets the default character shown instead of the password for password inputs.
+// A space can be used to hide the password length by not printing anything.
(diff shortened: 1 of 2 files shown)


AlecAivazis:
This seems good - mind adding a test case to [this file[(https://github.com/go-survey/survey/blob/master/tests/password.go)?

szh:
> This looks good - mind adding a test case to [this file](https://github.com/go-survey/survey/blob/master/tests/password.go)?

I don't see how to add a test case there since the prompt is the same. I think this would require some refactoring. Do you want me to attempt?

szh:
@AlecAivazis can we get this moving?

AlecAivazis:
Yea feel free to refactor the tests to verify this behavior

AlecAivazis:
Whoops! Wrong button 😅

PR #483, opened 2023-04-06
Fix bug when deal with input value of `uint`.

hustclf:
hi, guys,  there is a bug when parsing input unit value larger than 255.

here is a example to reproduce the bug.
```
package main

import (
)

func main() {
	var max uint
	if err := survey.AskOne(&survey.Input{
		Message: "Which is the unit value?",
	}, &max); err != nil {
		panic(err)
	}
	print(max)
}

```

image

And we should set `bitSize` to 0 when calling `parseUint`, which means let the operating system determine the appropriate `bitSize`.

### DIFF
--- a/core/write.go
+++ b/core/write.go
@@ -268,7 +268,7 @@ func copy(t reflect.Value, v reflect.Value) (err error) {
 			}
 		case reflect.Uint:
 			var val64 uint64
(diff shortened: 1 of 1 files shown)


AlecAivazis:
Awesome! Thanks for putting this together.

Do you mind adding a test that confirms the behavior?

PR #484, opened 2023-04-26
fixes #465

### LINKED ISSUE #465: Rename package to `github.com/go-survey/survey/v2` in `go.mod`

Executing `go get -u github.com/go-survey/survey/v2`

No error

**What did you see instead?**

```
go: downloading github.com/go-survey/survey/v2 v2.3.6              
go: github.com/go-survey/survey/v2@v2.3.6: parsing go.mod:           
        module declares its path as: github.com/AlecAivazis/survey/v2     
                but was required as: github.com/go-survey/survey/v2                                                                                                                           
```

The GitHub repository has been renamed, but not the package name in go.mod:

https://github.com/go-survey/survey/blob/93657ef69381dd1ffc7a4a9cfe5a2aefff4ca4ad/go.mod#L1

mfiumara:
This module should be renamed to reflect the url where it's located now.

### DIFF
--- a/confirm_test.go
+++ b/confirm_test.go
@@ -7,8 +7,8 @@ import (
 	"os"
 	"testing"
 
-	"github.com/AlecAivazis/survey/v2/core"
-	"github.com/AlecAivazis/survey/v2/terminal"
+	"github.com/go-survey/survey/v2/core"
+	"github.com/go-survey/survey/v2/terminal"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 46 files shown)


mfiumara:
@mislav / @AlecAivazis Could you provide some feedback on this change? It's not 100% clear for current users on how to start using this package as you still have to use the old URL in your Go project to be able to use the package.

I don't think this should break existing repositories using this package, correct?

AlecAivazis:
I think that's correct since GitHub will still proxy the requests but I'll wait for @mislav to comment

